PR #8, opened 2015-01-05
[WIP] StackPHP integration (fixes #5)

### LINKED ISSUE #5: StackPHP integration
As suggested with @frankdejonge, it might be good to add @stackphp support to Glide, since it's primarily an HTTP based library. Also, as per @alexbilbie's suggestion, the URL signing feature could be split out as a separate middleware as well, although I'm not exactly sure how that would work.

alexbilbie:


alexbilbie:
I think I'm going to alter (what is now called) ImageRequest to extend Symfony's Request class because it has some built in functionality that you've remade. It also is one less object that is created :)

reinink:
Yeah, I was wondering about just using the HttpFoundation's Request object instead. All I REALLY need is a way to hash it (to use as the filename), which is really a very simple [method](https://github.com/thephpleague/glide/blob/stackphp/src/ImageRequest.php#L84). The `getHash()` method is actually only ever used in the `Server` class, so we could even just move it there. Something like this:

```php
public function getHashedFilename(Request $request)
{
    $params = $request->query->all();

    unset($params['token']);

    return md5($request->getPathInfo().'?'.http_build_query($params));
}
```

reinink:
hey Alex, one question, I thought that the [constructor](https://github.com/thephpleague/glide/blob/master/src/Middlewares/ImageServer.php#L21) had to accept an instance of the `HttpKernelInterface` as the first parameter?

alexbilbie:
Sorry, my mistake. I will fix that

alexbilbie:
@reinink fixed in 0d72e458c92116dc0f75c617996826765302daff

PR #12, opened 2015-01-06
Using PHPUnit installed by composer.

h4cc:


reinink:
Interesting, good call. I think we may need to update our [skeleton](https://github.com/thephpleague/) project.

PR #23, opened 2015-01-13
Fixed version constraint

GrahamCampbell:


reinink:
Thanks Graham, what would I do without you?

GrahamCampbell:
:)

PR #27, opened 2015-01-23
Remove dev constraint

barryvdh:
Flysystem has been tagged.

Perhaps a new tag would be nice? Because the stable version still requires 0.5


GrahamCampbell:
:+1:

PR #28, opened 2015-01-23
Check cache freshness

barryvdh:
This validates the timestamp of the cache and source. If the cache isn't fresh anymore, the cache file is deleted.
Not sure if this should be optional or anything, because it will impact performance (every requests needs to make an API call).

Usecase is that the source image could be changed, but this will not be reflected in the Glide image.

For performance, a caching adapter for Flysystem could be suggested, which caches all calls (except for the content) so the impact wouldn't be very much. (http://flysystem.thephpleague.com/caching/)

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -280,9 +280,15 @@ class Server
     public function makeImage()
     {
         $request = $this->resolveRequestObject(func_get_args());
+        $cachePath = $this->getCachePath($request);
+        $sourcePath = $this->getSourcePath($request);
 
         if ($this->cacheFileExists($request) === true) {
-            return $request;
+            if ($this->cache->getTimestamp($cachePath) >= $this->source->getTimestamp($sourcePath)) {
+                return $request;
+            } else {
+                $this->cache->delete($cachePath);
+            }
         }
 
         if ($this->sourceFileExists($request) === false) {


reinink:
Very interesting idea!

You still have to deal with any browser level caching, which would require a filename change to force an update. My solution to this has simply been to append a revision query value to the URL. For example: `image.jpg?w=100&revision=2`. This essentially solves both problems, since the cached filename will change (assuming you're not stripping the revision param).

However, I still think this is worth considering. A request is already made each time to verify that the cache file exists, so if Flysystem caching was used, this wouldn't add any overhead there. However, it would require an additional check to see if the source file exists, so that would be additional overhead. Right now Glide doesn't check if the source file exists if a cache file was found.

So the question really is, should this be enabled by default, or be an option?

barryvdh:
Probably should be an option, as 99% of the time, the source doesn't actually change. And checking the source is probably more intensive then just the cache.
Using a timestamp would be an different possibility, so perhaps I could give the UrlBuilder the timestamp as an extra parameter, which would invalidate the cache automatically..

barryvdh:
Btw my interest in Glide was for this: https://github.com/barryvdh/elfinder-flysystem-driver
I integrated Flysystem with a filebrowser (for CMS editors etc) but didn't have a good way to show images/thumbnails, but with Glide I can just point to the api and generate urls for thumbs/images with the ImageBuilder :) And adding the timestamp to the query parameters does seem to mostly solve this issue :)

But it could be still interesting for some cases.

reinink:
I'm going to close this PR for now, since I think it needs to work, and also passing a query revision solves this problem for the most part.

PR #33, opened 2015-01-31
Decode URL entities in path

### LINKED ISSUE #32: Outputting image with a space in the filename fails when using a Request
When trying to load a file with a space (at a URL like `/img/my%20file.png`), it fails when using an instance of `Symfony\Component\HttpFoundation\Request`.

By failing, I mean we see a `NotFoundException: Could not find the image my%20file.png`.

``` php
public function serve($uri, \Illuminate\Http\Request $request)
{
  $server = ServerFactory::create([...]);

  // this fails:
  $server->outputImage($request);

  // this works: ($uri is 'my file.png' without the %20)
  $server->outputImage($uri, $request->query->all());
}
```

Both methods work when the filename has no spaces.

jasonvarga:
Simply wrapping the `$path` returned in `getSourcePath()` with `urldecode()` appears to fix #32.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -123,7 +123,7 @@ class Server
             $path = $this->sourcePathPrefix.'/'.$path;
         }
 
-        return urldecode($path);
+        return rawurldecode($path);
     }
 
     /**
--- a/src/Server.php
+++ b/src/Server.php
@@ -123,7 +123,7 @@ class Server
             $path = $this->sourcePathPrefix.'/'.$path;
         }
 
-        return $path;
+        return urldecode($path);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


reinink:
Thanks for working this issue out @jasonvarga! I'll get this tagged right away. You may find the very first item (at the bottom) of this [CHANGELOG](https://github.com/symfony/HttpFoundation/blob/8f39f4c3a2903883898d1026f781f0f42672f9bc/CHANGELOG.md) interesting. ;)

reinink:
Tagged: [0.3.1](https://github.com/thephpleague/glide/releases/tag/0.3.1)

jasonvarga:
Well would you look at that. Thanks!

PR #40, opened 2015-02-27
A bunch of stuff for Travis CI

duncan3dc:
* Use `composer update` to test against the highest dependencies
* Also test against the lowest supported dependencies
* Added testing on the upcoming php release (7.0)
* Use [container-based](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) travis builds 

Doing this uncovered an [issue](https://travis-ci.org/duncan3dc/glide/builds/52405993) in the dependencies. This looks like a problem in `intervention/image`, where they are misstating their lowest supported dependencies. Bumping glide's supported version to `~2.1` fixes that but I'm not sure whether you'd rather pursue a fix with intervention instead.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": ">=5.4",
-        "intervention/image": "~2.0",
+        "intervention/image": "~2.1",
         "league/flysystem": "~1.0",
         "symfony/http-foundation": "~2.3",
         "symfony/http-kernel": "~2.3"
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     },
     "require-dev": {
         "mockery/mockery": "~0.9",
+        "phpunit/php-token-stream": ">=1.3.0",
         "phpunit/phpunit": "~4.4"
     },
     "autoload": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,13 @@ php:
   - 7.0
   - hhvm
 
+env:
+  - COMPOSER_OPTS=""
+  - COMPOSER_OPTS="--prefer-lowest"
+
 before_script:
   - travis_retry composer self-update
-  - travis_retry composer install --no-interaction --prefer-source --dev
+  - travis_retry composer update $COMPOSER_OPTS
 
 script:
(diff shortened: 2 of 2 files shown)


duncan3dc:
It looks like there's another lowest dependency issue (only showing on hhvm), but I'm not sure which package is the culprit,

reinink:
Wow @duncan3dc, this is awesome! Thanks so much for putting this together. We just need to figure out that HHVM issue. I did a quick Google search, and [found this](http://stackoverflow.com/questions/24203912/hhvm-imagettftext-no-such-file-or-directory). Makes me wonder if this is related to GD. Although that doesn't make a ton of sense, since that test is mocked out pretty heavily. I'll have to do some digging.

duncan3dc:
Cool, I'm happy to help. I have an hhvm instance at home so I'll do some checking myself next time I get chance, if you don't beat me to the solution :)

duncan3dc:
It looks like the problem is `@runInSeparateProcess`, removing this allows the tests to pass.

Do you know why this particular test is run under this condition? The commit (dac2c96e) doesn't mention it.

duncan3dc:
Adding a requirement for `phpunit/php-token-stream` of at least version `1.3.0` seems to resolve this issue.

There are a few issues reporting this:
sebastianbergmann/phpunit#1090
sebastianbergmann/phpunit#1214
sebastianbergmann/phpunit#1407

And the resolution seems to just be "use a more up to date version". Are you happy to list the `php-token-stream` as a requirement for the `@runInSeparateProcess` feature?

reinink:
If I recall correctly, the `@runInSeparateProcess` option allows for output testing, so I suspect we'll need that still. Your solution here looks good to me!

reinink:
Thanks again for your work on this @duncan3dc, much appreciated!

duncan3dc:
Thanks for your work on producing this cool library :+1:

composer.json
```diff
@@ -22,13 +22,14 @@
...
         "php": ">=5.4",
-        "intervention/image": "~2.0",
+        "intervention/image": "~2.1",
         "league/flysystem": "~1.0",
         "symfony/http-foundation": "~2.3",
         "symfony/http-kernel": "~2.3"
     },
     "require-dev": {
         "mockery/mockery": "~0.9",
+        "phpunit/php-token-stream": ">=1.3.0",
```
GrahamCampbell: Don't use unbound version constraints.

reinink: Good point.

duncan3dc: It's not really a good point, as the dependency is only relevant for phpunit, which has a bounded version constraint, so we don't care what version of php-token-stream we get, as long as it's above 1.3.0, and compatible with our ~4.4 of phpunit, which composer takes care of

PR #45, opened 2015-03-09
Image is not resized

wakumaku:
When one of the dimensions requested is the same as the original the image is not resized.


reinink:
Wow, good find. Thanks for this! I'll merge it in and get it tagged soon. Little swamped right now.

PR #46, opened 2015-03-11
Add the ability to disable caching by passing false as the cache parameter

kevin-thackorie:
WIP. As per Jonathan's request, this potential fix for Glide's current inability to disable caching is being submitted as a pull request without unit test updates for the sake of implementation review.

These modifications enable the passing of boolean false as the value for the ‘cache’ parameter to the server factory. Doing so bypasses the cache writing process and serves the image from source.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 composer.lock
 vendor
+.DS_STORE
--- a/src/Server.php
+++ b/src/Server.php
@@ -29,7 +29,7 @@ class Server
 
     /**
      * The cache file system.
-     * @var FilesystemInterface
+     * @var FilesystemInterface|null
      */
     protected $cache;
 
@@ -53,11 +53,11 @@ class Server
 
     /**
      * Create Server instance.
-     * @param FilesystemInterface $source The source file system.
-     * @param FilesystemInterface $cache  The cache file system.
-     * @param ApiInterface        $api    The image manipulation API.
+     * @param FilesystemInterface      $source The source file system.
+     * @param FilesystemInterface|null $cache  The cache file system.
+     * @param ApiInterface             $api    The image manipulation API.
      */
-    public function __construct(FilesystemInterface $source, FilesystemInterface $cache, ApiInterface $api)
+    public function __construct(FilesystemInterface $source, $cache, ApiInterface $api)
     {
         $this->setSource($source);
         $this->setCache($cache);
@@ -141,16 +141,16 @@ class Server
 
     /**
      * Set the cache file system.
-     * @param FilesystemInterface $cache The cache file system.
+     * @param FilesystemInterface|null $cache The cache file system.
      */
-    public function setCache(FilesystemInterface $cache)
+    public function setCache($cache)
     {
         $this->cache = $cache;
     }
 
     /**
      * Get the cache file system.
-     * @return FilesystemInterface The cache file system.
+     * @return FilesystemInterface|null The cache file system.
      */
     public function getCache()
     {
@@ -200,6 +200,10 @@ class Server
      */
     public function cacheFileExists()
     {
+        if ($this->cache === null) {
+            return false;
+        }
+
         $request = $this->resolveRequestObject(func_get_args());
 
         return $this->cache->has($this->getCachePath($request));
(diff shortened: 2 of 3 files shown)


src/ServerFactory.php
```diff
@@ -79,6 +79,10 @@ public function getCache()
             $cache = $this->config['cache'];
         }
 
+        if ($cache === false) {
+            return null;
```
GrahamCampbell: just `return;`.

kevin-thackorie: Functionally, you're correct in that `return;` is equivalent to `return null;`, but I like to be explicit in circumstances such as this. It implies that the caller of this method will be specifically checking for or using the null return.

.gitignore
```diff
@@ -1,3 +1,4 @@
 build
 composer.lock
-vendor
\ No newline at end of file
+vendor
+.DS_STORE
```
GrahamCampbell: Nooo. This should never go on there.

reinink: Are you referring to `.DS_STORE`? Because I've already removed it locally. (I'm cleaning this up before tagging the next release.)

GrahamCampbell: Yeh. I've just sent a pull too, lol.

reinink: It's going to be removed twice. :)

GrahamCampbell: :ship:

kevin-thackorie: Why would one **not** want .DS_STORE files ignored? A .DS_STORE file "is a hidden file with a proprietary format created by OS X to store custom attributes of a folder such as the position of icons or the choice of a background image... these files were primarily used by the Finder [Mac application]"

I always add hidden temporary/application-irrelevant files to the .gitignore files of my projects.

reinink: It's generally discouraged, because where does it stop? You could end up with a VERY long .gitignore file in every project. Generally it's recommended to use a global .gitignore file isntead for this sort of stuff.

GrahamCampbell: > I always add hidden temporary/application-irrelevant files to the .gitignore files of my projects.

That's a bad practice. You should make use of global .gitignore files.

src/Server.php
```diff
@@ -53,11 +53,11 @@ class Server
...
      * Create Server instance.
-     * @param FilesystemInterface $source The source file system.
-     * @param FilesystemInterface $cache  The cache file system.
-     * @param ApiInterface        $api    The image manipulation API.
+     * @param FilesystemInterface      $source The source file system.
+     * @param FilesystemInterface|null $cache  The cache file system.
+     * @param ApiInterface             $api    The image manipulation API.
      */
-    public function __construct(FilesystemInterface $source, FilesystemInterface $cache, ApiInterface $api)
+    public function __construct(FilesystemInterface $source, $cache, ApiInterface $api)
```
reinink: I wonder if this is best. I really hate removing the `FilesystemInterface ` type-hinting, but I also don't want to change the parameter order because that would break backward compatibly.

We could do this:

```php
public function __construct(FilesystemInterface $source, FilesystemInterface $cache = null, ApiInterface $api = null)
```

Now clearly `ApiInterface` cannot be null, however if null is passed it will then get caught in the `setApi` method type-hinting. So, kind of weird, but maybe better than removing the `FilesystemInterface` type-hint?

Thoughts? /cc @GrahamCampbell

reinink: Or do we just do this, and who cares if PHP Code Sniffer baulks?

```php
public function __construct(FilesystemInterface $source, FilesystemInterface $cache = null, ApiInterface $api)
```

.gitignore
kevin-thackorie: Ahhh, I see what you mean. 

> You should make use of global .gitignore files.

I do, but omitting said files from the project-level .gitignore means that you now depend on other devs to have appropriately configured global .gitignore files. And, realistically, there are only 3 common files of that nature: thumbs.db (Windows), .DS_STORE (Mac), .swp (*nix).

This is a very small detail, though, and I'm good either way :)

src/Server.php
kevin-thackorie: +1 for the above suggestion of having the type hint on the $cache argument with a default value of null. I.e.

```php
public function __construct(FilesystemInterface $source, FilesystemInterface $cache = null, ApiInterface $api)
```

reinink: Only problem is this breaks [PSR-2 compliance](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#44-method-arguments), which might break build steps. This is annoying. Might have to just go back to the way you had it.

kevin-thackorie: Hmmm, I see what you mean: "Method arguments with default values MUST go at the end of the argument list." But changing the order of the arguments would break backwards compatibility.

Yeah, it's probably best to just leave the removal of the type hint for that argument.

PR #51, opened 2015-03-20
Remove .DS_STORE from .gitignore

GrahamCampbell:
This should never have been added. It belongs in your system .gitignore file.


reinink:
Thanks again for your all your help @GrahamCampbell!

GrahamCampbell:
np, thanks for an awesome package. :P

Issue #54, opened 2015-03-26
Image not resizing when cache disabled (dev-master)


BD-Rene:
When using the latest cache disabling stuff in dev-master, it seems that image resizing no longer works correctly. Haven't tried any other image manipulation, but something like this:

```
$server = League\Glide\ServerFactory::create([
'source' => dirname($_SERVER['SCRIPT_FILENAME']) . '/uploads/resources',
'cache' => false,
]);

$server->outputImage(thumbnail.jpg', [
'w'    => '200,
'fit'  => 'crop',
'crop' => 'center'
]);
```

... basically results in it outputting the original image without resizing or cropping.

Turning cache back on resolves this issue.

kevin-thackorie:
Confirmed. I am able to reproduce this issue and am investigating it now.

reinink:
This has been "fixed" in [0.3.5](https://github.com/thephpleague/glide/releases/tag/0.3.5), as the cache disabling now uses `sys_get_temp_dir()` when no cache is set.

However, we aware that caching disabling will be completely removed in the next version of Glide, in favour of a new in-memory adapter that handles this much nicer. See [here](https://github.com/thephpleague/glide/pull/56#issuecomment-112082800) for more information on that.

Issue #55, opened 2015-03-26
Impossible to overwrite ResponseFactory::create


lsv:
As ResponseFactory only streams the response (StreamedResponse), where I want is just to get a Response back, because im getting the images as base64 encoded instead of downloading them.

So I think ResponseFactory in Server should be overwriteable.

Maybe with

````
public function __construct(FilesystemInterface $source, FilesystemInterface $cache, ApiInterface $api, ResponseFactory $reponse = null)
{
    $this->setSource($source);
    $this->setCache($cache);
    $this->setApi($api);
    $this->setResponseFactory($response);
}
````

And now overwrite all the calls with

````
ResponseFactory::create(...)
````

to

````
$this->getResponseFactory()::create(...)
````

**Comment by reinink** [comment]
I'm actually planning on making the response factory an injectable dependency, just so folks can modify the response object as needed. So good call there!

However, in your particular situation, you shouldn't even be using responses, because you don't actually want to send a response with Glide, but rather you just want the manipulated image encoded base64. This can be done pretty easily, it just takes a little bit of extra code. Here's how:

```php
// Setup your Glide server
$glide = ServerFactory::create([
    // your configuration
]);

// Generate the manipulated image
$glide->makeImage($request);

// Get the path of maipulated image in the cache
$path = $glide->getCachePath($request);

// Get the content of the maipulated image from the cache
$source = $cache->read($path);

// Base64 encode your image (and even auto set the mime type)
$base64 = 'data:'.$cache->getMimetype($path).';base64,'.base64_encode($source);

// Use the Base64 encoded image in your app
echo '<img src="'.$base64.'">';
```

reinink:
The next iteration of Glide will have a `ResponseFactory` interface, which you can define within the `ServerFactory` options. Something like this:

```php
$glide = ServerFactory::create([
    'source' => $filesystem,
    'cache' => $cache,
    'response' => new MyResponseFactory(),
]);
```

If no `response` is set, it will just default to the `StreamedResponseFactory` class.

reinink:
Closing this ticket. Responses have been totally reworked for the next version, and this functionality has been merged into the master branch and will be tagged shortly when I release `v1.0`! Hopefully not long now.

Also, see the work-in-progress docs for this feature [here](http://glide.thephpleague.com/1.0/config/responses/).

PR #56, opened 2015-04-01
Fix bug in cache disabling feature that ignores manipulations

kevin-thackorie:
- Add the UncachedResponseFactory
- Serve the manipulated image from memory without expensive disk I/O

This fixes issue https://github.com/thephpleague/glide/issues/54

Note: ```\Finfo``` is used and was included in PHP >= 5.3

### DIFF
--- a/src/Http/UncachedResponseFactory.php
+++ b/src/Http/UncachedResponseFactory.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace League\Glide\Http;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+class UncachedResponseFactory
+{
+    /**
+     * The final image's binary string.
+     * @var string
+     */
+    protected $inMemoryImage;
+
+    /**
+     * The request object.
+     * @var Request
+     */
+    protected $request;
+
+    /**
+     * The timestamp representation of the source file's last modified date.
+     * @var integer
+     */
+    protected $sourceLastModified;
+
+    /**
+     * Create Output instance.
+     * @param string  $cache              The cache file system.
+     * @param Request $request            The request object.
+     * @param integer $sourceLastModified Timestamp
+     */
+    public function __construct($imageBinary, Request $request, $sourceLastModified)
+    {
+        $this->inMemoryImage      = $imageBinary;
+        $this->request            = $request;
+        $this->sourceLastModified = $sourceLastModified;
+    }
+
+    /**
+     * Get the streamed response.
+     * @return StreamedResponse The response object.
+     */
+    public function getResponse()
+    {
+        $response = new StreamedResponse();
+
+        $this->setHeaders($response);
+        $this->setContent($response);
+
+        return $response;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


kevin-thackorie:
@reinink This PR is needed to fix image manipulations with cache disabled in dev-master.

reinink:
Hey @kevin-thackorie, thanks so much for putting this together and for your ongoing help on this project! This seems like a pretty good solution, although I have a concern.

I wonder if this much extra code is "worth the effort", instead of just using the ```sys_get_temp_dir()``` as the cache when no cache is provided. It means having to write it to disk instead of serving it from memory, and I get that will have a slight performance hit, but how much? This will only happen once, and that's when the image is first generated, which is already going to be an expensive. I haven't done any testing on this either...just curious if it's worth it.

Also, is there any reason not to use the temp folder for this, other than for these potential performance reasons?

kevin-thackorie:
Hey @reinink, my pleasure!

Regarding your temp directory suggestion: I intentionally stored the binary of the manipulated image in memory prior to streaming it back to the user in the response because writing it to disk at any point would incur expensive disk I/O. This expense would not simply occur only once because, with the cache disabled, the entire process is invoked each time the image is requested.

reinink:
Well, yes and no. The idea behind disabling the cache is so another caching mechanism can be used instead, such as Varnish. So, subsequent requests should be handled by this alternative caching technique, and shouldn't hit Glide (unless that cache has expired).

Don't get me wrong, I'm not arguing that the in-memory solution isn't better, it's just a bunch of extra code that I need to review properly and unit test before tagging for production. So, until I have that figured out, I think using the system temp directory is an acceptable solution.

Cool? :)

kevin-thackorie:
True, in the case of an external caching layer like Varnish or Amazon CloudFront being placed in front of Glide, the disk I/O expense would only be incurred once.

I like that the solution in this PR works well whether or not the developer using Glide also has an external caching layer, but understand your point about the testing burden introduced by the UncachedResponseFactory.

Thus I defer to your judgment :)

reinink:
Hey @kevin-thackorie! I've finally worked out a really good solution here. We're [adding an in-memory adapter](https://github.com/twistor/flysystem-memory-adapter/issues/1) to [Flysystem](http://flysystem.thephpleague.com/), which does exactly this, without any extra code being required in Glide. With this approach, you can literally just do this:

```php
$glide = ServerFactory::create([
    'source' => $yourSource,
    'cache' => new Filesystem(new MemoryAdapter());
]);
```

I've already tested this, and it works perfect. In addition to this, in the next Glide release I'll be adding the ability to also override the `ReponseFactory` with your own factory. See #55 for more information on that.

I'm going to close this PR as a result. Let me know if you have any questions about this!

kevin-thackorie:
Hey @reinink, perfect! That is indeed a good solution as I felt Flysystem was lacking an in-memory adapter. 

This should work great, thanks!

reinink:
@kevin-thackorie Excellent, glad to hear you like this approach. You'll be happy to hear, that as of today, the in-memory adapter is a real thing!

http://flysystem.thephpleague.com/adapter/memory/

Issue #7, opened 2015-01-01
Cleaning up the cache


reinink:
Currently Glide does not do any cache cleanup, meaning it generates manipulated images and places them into the cache file system, but it never deletes them. It's possible that over time this cache gets large and contains many unused image manipulations.

The simple solution here is to simply delete the entire cache periodically, and on small projects this is probably fine. But on larger projects it could be problematic since all image manipulations will have to be regenerated.

It would probably be good if Glide came with a `CacheCleanup` helper. This cleanup tool could be run every time a request is received, or it could be setup with a scheduler (probably better). Something like:

```php
$cache = new Filesystem(new LocalAdapter('cache-folder'));
$cleanup = new CacheCleanup($cache)
$cleanup->expire(2592000); // 30 days
```

This approach would delete any manipulations older than 30 days. This can be determined easily use the `$filesystem->getTimestamp()` method.

However, this technique will not take into consideration whether they are being actively accessed or not. Using the last accessed date would be much more complicated since this information is not readily available. To do this you would need to create a data store, and save this information every time a manipulation is created. I feel this falls outside of the scope of this project. If this sort of cache cleanup is needed, developers are welcome to implement this themselves.

h4cc:
If there are multiple ways of implementing the CacheCleanup, it could be integrated by using the Strategy Pattern.
Some default strategy like `OlderThanCleanup` or `NoCleanup` can then be provided by glide.

reinink:
@h4cc That's probably a good idea actually.

I've thought of another way of doing this. Glide could keep an access log in the cache file system, which could be then used by a cleanup job later. This access log would provide access dates in addition to just creation dates.

barryvdh:
@reinink and what about when the source image changes?

For example, I have a filebrowser where I can resize/rotate etc some files. But changes aren't shown because the file is cached.
A solution could be checking the time the file has changed and invalidate the cache if needed, but this causes extra calls..

reinink:
I'm actually like the idea I'm seeing in PR #60.

Basically, we could allow folks to override how the cached filename is generated. This would allow you to group all manipulations for a specific source image into a specific folder. Then, if you delete that source folder, you can easily delete all the cached variations. Something like this:

```php
$server->setCachePathFilter(function (Request $request, $path) {
    // modify $path
    return $path;
});
``

At this point I'm just not sure a closure is the best way (it seems a little complicated), or if we just abstract the cache path generation to a new class, with a interface.

[...]

PR #60, opened 2015-04-24
Cache path filter closure for modifying the cache path relative to its self

AshleyDawson:
When caching to some local filesystems, it's necessary to implement a nested directory structure to get around maximum object limitations.

I've added a cache path filter (closure) allowing you to modify the cache path relative to it's self in some way. The example I've given in the docblock is prepending a nested directory structure - but I guess you could modify the path in any way using this filter.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -50,6 +50,12 @@ class Server
      * @var string
      */
     protected $baseUrl;
+    
+    /**
+     * Modify the cache path relative to its self
+     * @var \Closure
+     */
+    protected $cachePathFilter;
 
     /**
      * Create Server instance.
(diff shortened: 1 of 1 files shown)


reinink:
Hey @AshleyDawson! Very sorry for the lack of response on this, I've been buried with life recently. I actually like the idea you present here, although I'm not convinced on the design quite yet. I'd like to do some testing first.

There's a discussion that's related to this here: #7.

reinink:
Thanks again for submitting this, but I think I'm going to close this in favour of the solution suggested [here](https://github.com/thephpleague/glide/issues/7#issuecomment-112161637). I think it's just a little simpler, and accomplishes the same thing. However, I appreciate your contribution here, as it led me to that solution. Thanks!!!

PR #78, opened 2015-06-25
Add fit=fill feature for resize

jcarouth:
When resizing an image which is smaller than the dimensions you want to
resize it to in one or both directions, the fit=fill option allows you
to specify a fill color (when combined with the Background manipulator)
to fill the extra canvas space.

This is accomplished using a query like `?w=300&fit=fill&bg=000` to resize
the image to 300 pixels wide while filling any new canvas space with
black.

### DIFF
--- a/src/Manipulators/Size.php
+++ b/src/Manipulators/Size.php
@@ -104,7 +104,7 @@ class Size extends BaseManipulator
      */
     public function getFit()
     {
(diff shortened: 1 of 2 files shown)


reinink:
Nice work Jeff, thanks for adding this!

jcarouth:
:+1:

PR #89, opened 2015-08-07
remove Bad Apostrophe from README.md

curtisgibby:
A possessive belonging to an "it" doesn't need an apostrophe. Don't believe me? Ask [the Oatmeal](http://theoatmeal.com/comics/apostrophe) (look for the velociraptor)!


reinink:
Good catch! And I do believe you! :)

PR #95, opened 2015-08-13
Added a ColorSpace Manipulator

joshrhykerd:
I needed the ability to change the color space when working with the Imagick driver. Prior to being able to change the color space my Glide generated images were being generated with incorrect colors. With this update a user can use the "cs" option to the specify one of the defined colorspaces (rgb, srgb, cmyk). I wasn't sure if this was possible with the GD driver so my implementation only works with Imagick. Intervention also does not implement any commands to handle this so I had to use the Imagick driver directly.

### DIFF
--- a/src/Manipulators/ColorSpace.php
+++ b/src/Manipulators/ColorSpace.php
@@ -29,7 +29,7 @@ class ColorSpace extends BaseManipulator
 
         $this->colorSpaces = [
             'rgb' => \Imagick::COLORSPACE_RGB,
-            'srgb' => \Imagick::COLORSPACE_RGB,
+            'srgb' => \Imagick::COLORSPACE_SRGB,
             'cmyk' => \Imagick::COLORSPACE_CMYK
         ];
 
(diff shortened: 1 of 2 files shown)


reinink:
Hey Josh! Thanks so much for your interest in Glide, and for contributing this back. It's always very much appreciated. However, I'm not sure at this point that 1) I want an ImageMagick only manipulator, and 2) see a big enough use-case for setting the ColorSpace.

I think it's awesome that you were able to solve this for your particular use-case, and in fact I built Glide to allow this sort of customization. However, I'm not sure this belongs to the core package at this time.

I hope that makes sense, and either way, thanks so much!

joshrhykerd:
Thanks for your reply @reinink. I understand and still very much appreciate your contributing Glide to the community. I know it has been very useful for me in my team in several applications!

reinink:
@joshrhykerd Awesome to hear, thanks for the positive feedback! :+1:

PR #104, opened 2015-10-18
Circle manipulator

pascalbaljet:
I needed to draw circles the same way you would use border-radius with CSS. This commit adds the 'circle' parameter :)

### DIFF
--- a/src/Api/ApiFactory.php
+++ b/src/Api/ApiFactory.php
@@ -6,6 +6,7 @@ use Intervention\Image\ImageManager;
 use League\Glide\Api\Manipulator\Blur;
 use League\Glide\Api\Manipulator\Brightness;
 use League\Glide\Api\Manipulator\Contrast;
+use League\Glide\Api\Manipulator\Circle;
 use League\Glide\Api\Manipulator\Filter;
 use League\Glide\Api\Manipulator\Gamma;
 use League\Glide\Api\Manipulator\Orientation;
@@ -77,6 +78,7 @@ class ApiFactory
             new Orientation(),
             new Rectangle(),
             new Size($maxImageSize),
+            new Circle($maxImageSize),
             new Brightness(),
             new Contrast(),
             new Gamma(),
--- a/src/Api/Manipulator/Circle.php
+++ b/src/Api/Manipulator/Circle.php
@@ -1,0 +1,154 @@
+<?php
+
+namespace League\Glide\Api\Manipulator;
+
+use Intervention\Image\Image;
+use Intervention\Image\ImageManager;
+use Symfony\Component\HttpFoundation\Request;
+
+class Circle implements ManipulatorInterface
+{
+    /**
+     * Maximum image size in pixels.
+     * @var int|null
+     */
+    protected $maxImageSize;
+
+    /**
+     * Create Size instance.
+     * @param int|null $maxImageSize Maximum image size in pixels.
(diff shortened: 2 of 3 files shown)


reinink:
Decided against this?

reinink:
I was going to suggest that this might be better done with border-radius. Less problematic, and border-radius is widely supported now.

kleisauke:
@reinink I've made something similar but also support for the following shapes:
* Ellipse
* Star (3, 4 and 5 point star)
* Triangle
* Square
* Pentagon

The last 4 options are drawn with the [polygon option](http://image.intervention.io/api/polygon).

Source can be found here: https://gist.github.com/kleisauke/26a4f9d31469e9c0470b
(Not yet finished because the star, triangle, square and pentagon option are always the same size, and I think there are to many if else statements...)

Example output if we are using [this image](https://upload.wikimedia.org/wikipedia/commons/e/e8/Lichtenstein.jpg) (?w=160&star):
Star

Let me know if this is applicable for Glide and whether it is worth a pull request.

reinink:
@kleisauke Hey, thanks for sharing this! Just curious, what sort of applications have you used this? This is pretty cool, don't get me wrong, I just wonder how realistic it is that someone would actually use this feature.

kleisauke:
@reinink We intend to use this in the future for our [image cache & resize proxy](https://github.com/andrieslouw/imagesweserv). I'm also not sure if someone actually would use this feature, but it's worth a try.

vesper8:
@kleisauke Hello! This is exactly what I need for my project... don't want to use border-radius.. want to pre-transform into a shape and cache the image

Unfortunately I see your gist is no longer available.

I took a quick look at your project and found this class https://github.com/andrieslouw/imagesweserv/blob/3.x/src/Manipulators/Shape.php which seems to be related, but it's not clear whether you make use of Glide at all and I don't know how to implement this with Glide.

Any chance you could restore the gist? Or provide a hint as to how to implement your new class with Glide please?

Many thanks!

kleisauke:
@vesper8 Here you can find the shape manipulator that should be compatible with Glide:
https://github.com/andrieslouw/imagesweserv/blob/78d8b32513eb630b57b0615977acc5464777630b/src/Manipulators/Shape.php

From version 3.x we dropped [Intervention Image](https://github.com/Intervention/image) in favor of [php-vips](https://github.com/jcupitt/php-vips) because resizing an image with [libvips](https://github.com/jcupitt/libvips) is typically 4x-5x faster than using the quickest ImageMagick.

vesper8:
Thanks for taking the time! Would you mind also telling me how I can use this custom manipulator? I've tried finding out more about custom manipulators in the Glide docs and issues and this issue is actually one of the only things that comes up.. there's like no help that I could find.. would really save me some time if you could write a few more lines on the subject :)

thanks!

kleisauke:
@vesper8 I've created an example project with a custom Glide manipulator, see the attachment.
1. Install dependencies with:
```sh
composer install
```

2. Configure the server to redirect all requests to `index.php` (see `glide.nginxconf` for an example nginx config).

3. Open `localhost` in your browser (or wherever you've hosted the website). You'll see something like this:
shape-manipulator

Attachment: Example project as zip file.

[custom-glide-manipulator.zip](https://github.com/thephpleague/glide/files/1099688/custom-glide-manipulator.zip)

PR #113, opened 2015-12-21
Added Dockerfile to jump start the curious

fschaper:
Added a Dockerfile so that people interested in glide can take a quick peek.


reinink:
Hmm, this is interesting. Thanks so much for your interest in Glide, and for contributing to this project. I'm not certain I want to include this yet. My concern is where does it stop? Do I offer a Vagrant file in the future as well? I wonder if this would be better in a guide/example app.

fschaper:
Entirely your decision.

If you were to offer an example for people to conveniently try out glide and play with the API, an example repository would make sense to me. If you were to provide a reference installation that could be used as a single stop solution, it could go either way for me.

Since your project pretty much is a complete solution in and of itself I see the value in having the option to deploy a reference installation  - I at least would be interested.

reinink:
So, it seems to me that this PR actually deletes all the libraries code...that's a big problem.

However, while I appreciate the interest in this library, and your efforts, I'm thinking that I don't want such a specific solution included in this project. I couldn't see myself adding an example vagrant file either.

I do think there is value in a complete code example, I think I'll add that!

fschaper:
Hello reinik,
after our short discussion concerning the merit of having a dedicated Dockerfile as part of the glide project, I removed all the glide project components and moved the Dockerfile to it's own repository under https://github.com/fschaper/glide-dockerized in the hope that it will be useful for somebody.

Cheers,
Florian

reinink:
Very cool! I've tweeted about this: https://twitter.com/reinink/status/699667816056619008

PR #124, opened 2016-02-16
Improve Glide UrlBuilder class

nyamsprod:
The class now uses internally [League Uri modifiers](http://uri.thephpleague.com/uri/manipulation/) to manipulate the HTTP Uri

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "intervention/image": "^2.1",
         "league/flysystem": "^1.0",
         "php": "^5.4 | ^7.0",
-        "psr/http-message": "^1.0"
+        "psr/http-message": "^1.0",
+        "league/uri": "^4.0"
     },
     "require-dev": {
         "mockery/mockery": "~0.9",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "db8e18e9e990b741953dd20a81b66e00",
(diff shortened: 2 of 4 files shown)


ADmad:
Unless there is a problem with existing URL manipulation code why add extra dependencies :)

Also league/uri requires PHP >= 5.5 while glide currently supports PHP >= 5.4. leauge/uri also requires `intl` extension which often isn't available on many systems or they have outdated ICU libs.

reinink:
@ADmad Yeah, those  are basically the reasons I'm not using the Uri library yet. It would be very handy, since the URL stuff is a pain. Maybe in the next release of Glide.

ADmad:
@reinink Gotcha :)

reinink:
Just revisiting this. I'm now working on [Glide 2.0](https://github.com/thephpleague/glide/issues/170). I'm thinking this library is probably still not necessary, only because the URL generation in Glide 2.0 is extremely simple: https://github.com/thephpleague/glide/blob/2.0/src/Image.php#L140-L145

@nyamsprod, can you see any issues with that current URL builder implementation?

Don't get me wrong, I love the `thephpleague/uri` package. I just think it's a little overkill this for this use-case. Plus the extra dependences would be nice to avoid, especially the `intl` extension.

nyamsprod:
@reinink To be fair I had completely forgot about this PR. So after diving into it again, from what I can tell, `Server::getBaseUrl` and `Server::getCacheUrl` are always return a string path and nothing else then yes, you don't need the league URI. **TL;DR** If you want to close this PR, please do so :+1:

reinink:
@nyamsprod Great, thanks! 👌🏻

PR #126, opened 2016-02-28
Add ability to set the opacity of the watermark

bonzai:


reinink:
I like this idea. I just need to find a moment to review this closer, but my initial reaction is I think this will be merged in. Thanks for contributing!

reinink:
Tagged: https://github.com/thephpleague/glide/releases/tag/1.2.0

Documented: http://glide.thephpleague.com/1.0/api/watermarks/#alpha-markalpha

PR #134, opened 2016-05-27
Fixed jpeg progressive and make sure temp image is deleted

thomaspicquet:
I run glide on a very large site, lots of traffic and we had issues with the temp image filling up the server just because one image was broken, also the part setting the jpeg image in a white image was breaking the progressive jpeg format. I don't see the utility of it so I removed it. I can put it back for jpeg only, it just doesn't work for progressive jpeg

### DIFF
--- a/src/Manipulators/Encode.php
+++ b/src/Manipulators/Encode.php
@@ -25,12 +25,6 @@ class Encode extends BaseManipulator
             $format = 'jpg';
         }
 
-        if ($format === 'jpg') {
-            $image = $image->getDriver()
-                           ->newImage($image->width(), $image->height(), '#fff')
-                           ->insert($image, 'top-left', 0, 0);
-        }
-
         return $image->encode($format, $quality);
     }
 
--- a/src/Manipulators/Size.php
+++ b/src/Manipulators/Size.php
@@ -305,7 +305,7 @@ class Size extends BaseManipulator
         list($resize_width, $resize_height) = $this->resolveCropResizeDimensions($image, $width, $height);
 
         $image->resize($resize_width, $resize_height, function ($constraint) {
-            $constraint->aspectRatio();
+            //$constraint->aspectRatio();
         });
 
         list($offset_x, $offset_y) = $this->resolveCropOffset($image, $width, $height);
--- a/src/Server.php
+++ b/src/Server.php
@@ -480,7 +480,7 @@ class Server
         // We need to write the image to the local disk before
         // doing any manipulations. This is because EXIF data
         // can only be read from an actual file.
-        $tmp = tempnam(sys_get_temp_dir(), 'Glide');
+        $tmp = tempnam(sys_get_temp_dir(), 'Glide'.rand(0,10));
 
         if (file_put_contents($tmp, $source) === false) {
             throw new FilesystemException(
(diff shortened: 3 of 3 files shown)


reinink:
Interesting, so somehow an exception is being thrown, and the temp image isn't being deleted? Do you know at what point the exception/error is occurring?

reinink:
The reason a white background is being put on jpegs is because by default the background is black, which isn't a great default on the web. So, are you saying that you're using the `pjpg` option, and it's causing the images to no longer be progressive?

thomaspicquet:
@reinink from what I see in my logs there's an exception being thrown at some point for a division by 0 for a resize, in that case the temp image doesnt get deleted.

As for the pjpg, when you add the white background, the returned image isnt progressive anymore. I can put it back for jpg only if you'd like?

reinink:
Okay thanks, I'll try to figure out the error, and why the temp file isn't being deleted. But your solution looks pretty good.

As for the jpeg stuff, let me also do some testing here. I would think that we'd want the white background on progressive jpeg's as well, since they too would have black backgrounds by default in the event of a source image with transparency. I'll have to do some more digging here, because I'm not sure this solution is entirely the right fix.

thomaspicquet:
sounds good, thanks!

reinink:
For sure, happy to help! I'm curious, you say you're using Glide on a large project with lots of traffic. Can you give me any sense of these numbers? I often have folks ask me if Glide can be used in this sort of way, so having some concrete examples would be great! (even just numbers, you don't have to share the site)

thomaspicquet:
There's currently around a thousand requests a minute on Glide and its behind a CDN where images are cached there for a year (around 6mil hits a day on the images), we pass a version of the image in the query to invalidate the cache when we have to.
We ran into issues using local cache with Glide as the filesystem couldn't keep up, right now we use the memory adapter and it fixed the issue. We used Silex and we also had to update the symfony response to be able to set a custom age for the HTTP cache, I believe it was hardcoded to 31 days.

reinink:
Yeah, with that design there really isn't any reason to even cache the images with Glide, since you're caching them with the CDN. Very cool! Thanks for sharing.

francescolaffi:
I also run in some problems with pjpg, it seems they are never generated as progressive... looking at the source and this PR I'd say the problem in `Encode` is that `interlace()` is called on an image object that is immediately replaced with a new, not interlaced, one.

Let me know if its useful if I open a different PR fixing just the instructions order, or you prefer handling it with this one

reinink:
Oh wow, yep, that it. Thanks @francescolaffi, I'll get this fixed.

ADmad:
I too noticed the lack of progressive images :)

ADmad:
A polite ping. Would be great to have the issue with progressive jpg fixed.

juukie:
Any news on this PR?

reinink:
Okay, the progressive jpeg encoding has been fixed here: f44f0efee7d80e5b351055e1ffbfa4d1368a04a6

I'll get that tagged right away.

reinink:
@thomaspicquet Did you happen to have the broken image around still? I'd love to test this myself and try to fix it.

thomaspicquet:
@reinink I don't unfortunately

reinink:
@thomaspicquet Any chance you could dig up the exception/error you got?

trajedy:
This is still an issue with the tempfile not being unlinked when the following exceptions are thrown. The /tmp folder can grow quite considerably with sites that generate a lot of images.

https://github.com/thephpleague/glide/blob/master/src/Server.php#L520
https://github.com/thephpleague/glide/blob/master/src/Server.php#L532

PR #137, opened 2016-06-07
Make sure stream is seekable before using `rewind()`

jackmu95:
This commit fixes possible `rewind(): stream does not support seeking` errors.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -439,7 +439,9 @@ class Server
 
         $stream = $this->cache->readStream($path);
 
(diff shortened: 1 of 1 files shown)


TheHiddenHaku:
We have the same problem and this fixes it.
Please, merge!

reinink:
Thanks for providing this fix @jackmu95!

reinink:
Merged and [tagged](https://github.com/thephpleague/glide/releases/tag/1.0.1).

PR #152, opened 2016-08-19
Add support for caching with file extensions

jasonvarga:
Hey there.

I'd like to be able to add file extensions to the cached images so they can be served directly.

In our app (Statamic, in this case), we'll have an option to pre-generate the images and use this approach. Obviously we'd lose the dynamic URL based nature of Glide, but this would be a switch people could flick to increase performance.

On heavier sites, loading an image directly is considerably faster than bootstrapping the app.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -43,6 +43,12 @@ class Server
     protected $groupCacheInFolders = true;
 
     /**
+     * Whether to cache with file extensions.
+     * @var bool
+     */
+    protected $cacheWithFileExtensions = false;
+
+    /**
      * Image manipulation API.
      * @var ApiInterface
      */
@@ -229,6 +235,24 @@ class Server
     }
 
     /**
+     * Set the cache with file extensions setting.
+     * @param bool $groupCacheInFolders Whether to cache with file extensions.
+     */
+    public function setCacheWithFileExtensions($cacheWithFileExtensions)
+    {
+        $this->cacheWithFileExtensions = $cacheWithFileExtensions;
+    }
+
+    /**
+     * Get the cache with file extensions setting.
+     * @return bool Whether to cache with file extensions.
+     */
+    public function getCacheWithFileExtensions()
+    {
+        return $this->cacheWithFileExtensions;
+    }
+
+    /**
      * Get cache path.
      * @param  string $path   Image path.
      * @param  array  $params Image manipulation params.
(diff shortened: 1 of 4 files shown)


reinink:
So the only issue I see here is when you specify a different file [encoding](http://glide.thephpleague.com/1.0/api/encode/) in your image manipulations (`?fm=jpg`). The source file may be a `png`, but the outputted image may be a `jpg`.

I think we could check for this though. We'd need to check the `$params` array, as well as any [defaults or presets](http://glide.thephpleague.com/1.0/config/defaults-and-presets/). Something like this:

1. Use `fm` if present in current `$params` list
2. Use `fm` if a preset (`p`) is being used and `fm` is set for that preset (`$this-> presets`)
3. Use `fm` if set in the defaults (`$this-> defaults`)
4. Use the format of the source file (`pathinfo($path)['extension']`)

jasonvarga:
Ah yes of course. Forgot about `fm`. I'll get that sorted.

reinink:
Thanks man! I'd help you but I'm out of time for this week.

jasonvarga:
No worries. I just pushed up the changes to support that.

src/Server.php
```diff
@@ -248,13 +272,18 @@ public function getCachePath($path, array $params = [])
...
+        $cachedPath = $this->groupCacheInFolders ? $sourcePath.'/'.$md5 : $md5;
 
         if ($this->cachePathPrefix) {
-            $path = $this->cachePathPrefix.'/'.$path;
+            $cachedPath = $this->cachePathPrefix.'/'.$cachedPath;
+        }
+        
+        if ($this->cacheWithFileExtensions) {
+            $ext = isset($params['fm']) ? $params['fm'] : pathinfo($path)['extension'];
+            $cachedPath .= '.'.$ext;
```
reinink: Could we make this a single line?

```php
$cachedPath .= isset($params['fm']) ? $params['fm'] : pathinfo($path)['extension'];
```

jasonvarga: Absolutely. Done.

reinink:
This is looking great. Next question. Is this `1.0.2` or `1.1.0`? It's not a bug patch, so technically by semver this is a minor version adding functionality in a backwards-compatible manner. So let's maybe do `1.1.0`.

jasonvarga:
Sounds good to me!

reinink:
Tagged: https://github.com/thephpleague/glide/releases/tag/1.1.0

jasonvarga:
Thanks!

PR #158, opened 2016-10-16
[Travis] Test on PHP 7.1

duncan3dc:


reinink:
Good call, thanks @duncan3dc!

PR #163, opened 2016-11-15
Documentation: Add Request object to LaravelResponseFactory

juukie:
The `SymfonyResponseFactory` can be constructed with an instance of the `Symfony\Component\HttpFoundation\Request`.
This will set the `last-modified` header based on the timestamp of the cached image.

Perhaps this should be added to the Symfony integration page as well, but I'm not sure how to pass in the Request object.


reinink:
Thanks @juukie!

PR #165, opened 2016-12-07
Makes temp dir configurable

### LINKED ISSUE #156: Server temp dir should be configurable
At the moment the glide server is using sys_get_temp_dir to be able to do image manipulations. 

``` php
        // We need to write the image to the local disk before
        // doing any manipulations. This is because EXIF data
        // can only be read from an actual file.
        $tmp = tempnam(sys_get_temp_dir(), 'Glide');

        if (file_put_contents($tmp, $source) === false) {
            throw new FilesystemException(
                'Unable to write temp file for `'.$sourcePath.'`.'
            );
        }
```

Our server doesn't allow vhosts to access the /tmp folder. It would be nice if this path was configurable. /tmp could still be a fallback so we don't have any bc breaks. Option should be a string so the user can only define a local path.

jaapio:
These changes make it posible to set the temp dir used by
the glide server to store images for processing.

Since this must be a local path we check if the path exists and is
writable by php. Otherwise an exception is thrown.

By default the tempDir property is set to `sys_get_temp_dir()` so no
changes are required for existing systems.

Fixes #156

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -79,6 +79,12 @@ class Server
     protected $presets = [];
 
     /**
+     * Path to store temp files.
+     * @var string
+     */
+    protected $tempDir;
+
+    /**
      * Create Server instance.
      * @param FilesystemInterface $source Source file system.
      * @param FilesystemInterface $cache  Cache file system.
@@ -89,6 +95,7 @@ class Server
         $this->setSource($source);
         $this->setCache($cache);
         $this->setApi($api);
+        $this->tempDir = sys_get_temp_dir();
     }
 
     /**
@@ -510,7 +517,7 @@ class Server
         // We need to write the image to the local disk before
         // doing any manipulations. This is because EXIF data
         // can only be read from an actual file.
-        $tmp = tempnam(sys_get_temp_dir(), 'Glide');
+        $tmp = tempnam($this->tempDir, 'Glide');
 
         if (file_put_contents($tmp, $source) === false) {
             throw new FilesystemException(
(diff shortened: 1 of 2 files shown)


reinink:
@jaapio Out of curiosity, what host were you having these issues with? I honestly thought there was a standard way of getting the available temp directory on any PHP based web server.

jaapio:
We are using mod-ruid to let the application run as the owner of the files. This user has a base_dir restriction so php is not allowed to access /tmp

anhphamt:
I also get this issue when working in my local environment OSX. It's fine when working with Host. 

Can we merge this pull request so that dev can adjust the tmp dir dynamically. 

Thank you

jorenvh1:
Can anyone explain me how I would go about using this fix to set the tmp folder in the root of my laravel project for example?

hykw:
We've also got the issue. Could you merge this pr, please?

piotr-cz:
I wonder why Travis builds failed here.
Job logs are not available anymore.

piotr-cz:
Nasty workaround:
Include file which defines `sys_get_temp_dir` function in `League\Glide` namespace, so it's used instead of native function.

```php
<?php
namespace League\Glide {
    /**
     * Monkey patch temp dir on shared hosting (open_base conflict)
     * @return string
     */
    function sys_get_temp_dir(): string
    {
        $nativeSysTempDir = \sys_get_temp_dir();

        // Your detection mechanism
        if (strpos($nativeSysTempDir, '/home/fcgid/') !== 0) {
            return $nativeSysTempDir;
        }

        return ini_get('upload_tmp_dir');
    }
}
```

tgalopin:
This sould be fixed by https://github.com/thephpleague/glide/pull/304, can anyone confirm?

PR #171, opened 2017-01-16
Fix source being trimmed when it begins with the baseUrl

jasonvarga:
I ran into a bug where if a path started with the baseUrl, it would get trimmed off incorrectly.

For example: if your `baseUrl` is `img`, and you do `$server->getSourcePath('imgur.jpg')` or `'img-0001.jpg` or something, it would return `ur.jpg` and `-0001` respectively.

I believe this PR fixes it by making sure there's a slash in the substring.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -136,9 +136,11 @@ class Server
     public function getSourcePath($path)
     {
         $path = trim($path, '/');
+        
+        $baseUrl = $this->baseUrl.'/';
 
-        if (substr($path, 0, strlen($this->baseUrl)) === $this->baseUrl) {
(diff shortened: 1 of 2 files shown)


reinink:
Wow, how on earth have we not run into this problem earlier? This solution looks good to me. I'll probably just merge you new `testGetsSourcePathWithBaseUrl()` test with the existing `testGetSourcePathWithBaseUrl()` test.

Thanks so much Jason!

reinink:
Tagged: [1.2.1](https://github.com/thephpleague/glide/releases/tag/1.2.1)

jasonvarga:
Derp. I think I renamed that method at the last moment. Thanks!

PR #187, opened 2017-04-23
Update Server.php

leonardyhuang:
Please check this out..


reinink:
Not sure what this does...seems like it would break Glide.

PR #190, opened 2017-04-30
add filter invert

wbadrh:
add invert filter
http://image.intervention.io/api/invert


reinink:
This seems well implemented. My only question, is there actually a valid use-case for this? Does anyone actually ever use invert? :)

wbadrh:
I want to implement every option in Intervention Image. They are available anyway.

Examples:
- invert alpha mask/channel
- easily place b&w jpeg logo on black or white background

reinink:
While I very much appreciate your interest in the library, I think I want to keep it focused on practical every day manipulations.

PR #194, opened 2017-05-09
Save pjpg as jpg when including file extensions

jasonvarga:
When you enable `setCacheWithFileExtensions` and use `fm=pjpg`, the file was being saved as `.pjpg`. Now it'll use `jpg`.

I'm not sure if this is the best way to fix it.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -281,7 +281,9 @@ class Server
         }
         
         if ($this->cacheWithFileExtensions) {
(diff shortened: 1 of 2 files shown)


reinink:
Thanks @jasonvarga, you've been awesome as always with this library! The `pjpg` format has created way too many bugs in this library. Fortunately that goes away in V2, which has simplified a bunch of things.

reinink:
Tagged: https://github.com/thephpleague/glide/releases/tag/1.2.2

Issue #118, opened 2016-01-11
Auto format support (for browsers with WebP and JPEG XR support) 


ifox:
Hello,

Both Cloudinary and Imgix have an auto format option that does automatic content negotiation. 
Is it in Glide's roadmap ?

reinink:
At this point no, mostly because I know very little about this. Can you tell me more about it?

ifox:
Sure, here are some resources from Imgix blog:

http://blog.imgix.com/2015/10/13/next-generation-responsive-images-with-client.html
http://blog.imgix.com/2014/08/20/webp-jpeg-xr-progressive-jpg-support-w-auto.html

User agent detection is ugly but Client Hints is a viable option, Chrome only though.

Another approach using Service Workers: http://www.deanhume.com/Home/BlogPost/service-workers--dynamic-responsive-images-using-webp-images/10132

That would solve our issue on the client side but Glide does not seem to provide WebP format because of Intervention Image not supporting it (https://github.com/Intervention/image/issues/233), right ?

jsor:
I just stumbled over this issue while researching [Client Hints](https://httpwg.github.io/http-extensions/client-hints.html). Here are few points to consider if Glide wants to support Client Hints:

1. Since Client Hints are sent as HTTP headers, Glide must support some kind of Request handling (maybe similar to the Reponses).
2. If the DPR header is used, the reponse [must include a Content-DPR header](https://httpwg.github.io/http-extensions/client-hints.html#rfc.section.3.1), i think this is also not possible at the moment. This header should probably be set from inside Size manipulator.
3. Since Glide sends far future expiration and cache-control headers, also a [Vary header must be set](https://httpwg.github.io/http-extensions/client-hints.html#rfc.section.2.2.2).
4. If a [Width header](https://httpwg.github.io/http-extensions/client-hints.html#rfc.section.4) overwrites the `w` query parameter and a `h`parameter is present, the value of `h` must be recalculated.

At the current state, i think it is not possible to support Client Hints easily. There must be some kind of middleware system to pre-process requests and responses and the parameters passed to the Glide server.

I've hacked a [little package](https://github.com/jsor/http-client-hints) together to help enabling Client Hints in Glide (though it's still a little rough at the moment).
It contains a `Resolver` class which can used like:

```php
$resolved = (new Jsor\HttpClientHints\Resolver())
    ->withWhitelist([
        // Process only Width and DPR headers
        'Width',
        'DPR',
    ])
    ->withMapping([
[...]

PR #202, opened 2017-07-07
Adds Support For WebP images

cdowdy:
This adds support for webp images.  

Intervention now has support for it
https://github.com/Intervention/image/commit/dccbee2e9dbde41b8b82deadd38fa574fc853689  
https://github.com/Intervention/image/commit/baeeac5125b7015f3fad0f9333b5e33585430f38  
https://github.com/Intervention/image/commit/961c58a9951acd688b9a2d17ff24ac9bcb199f84

I've added some tests for webp to EncodeTest.php but since I'm not to familiar with it I don't know if they're  right/good etc :)  

the docs (http://glide.thephpleague.com/1.0/api/encode/) will need a corresponding update as well  

related: https://github.com/thephpleague/glide/issues/118

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "intervention/image": "^2.1",
+        "intervention/image": "^2.4",
         "league/flysystem": "^1.0",
         "php": "^5.4 | ^7.0",
         "psr/http-message": "^1.0"
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "d6ed74d992f9e0130fe60f320e27e369",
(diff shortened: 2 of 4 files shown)


reinink:
Hey thanks for your work on this! I'd love to add webp support to Glide. I'll need to take a closer look at this. Can you fix the formatting and tests?

Also, from what I understand, webp can be either lossless or lossy. Does Intervention support a way to set which one you want?

cdowdy:
I don't know why the test is failing for 5.5/5/6 (https://travis-ci.org/thephpleague/glide/jobs/251241654#L259)  since GD has ``imagewebp`` support http://php.net/manual/en/function.imagewebp.php  


> Also, from what I understand, webp can be either lossless or lossy. Does Intervention support a way to set which one you want?

Not that I know of. From looking at it quickly it looks like they took the same route they took with PNG compression and just used whatever was the default of the driver being used. This though is an assumption

reinink:
Support for WebP has been added to Glide via another PR. Sorry for the delay on this @cdowdy, and thank you so much for submitting this either way! 👏

PR #208, opened 2017-10-14
Use namespaced TestCase

ZeljkaBlazek:
This PR resolves:
* tests now extend from namespaced TestCase

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - travis_retry composer self-update
 
 install:
-  - travis_retry composer update --no-interaction --prefer-source
+  - travis_retry composer install --no-interaction --prefer-source
 
 script:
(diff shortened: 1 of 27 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ before_install:
   - travis_retry composer self-update
 
 install:
-  - travis_retry composer update --no-interaction --prefer-source
+  - travis_retry composer install --no-interaction --prefer-source
```
GrahamCampbell: Actually, install is wrong, update is correct.

GrahamCampbell: It's only for BC that install works when there's no composer.lock present.

reinink:
Closing this, given the issues, and I just don't see this as a huge issue right now. Thanks either way! I appreciate your interest in the library.

PR #212, opened 2017-11-14
Number format of alpha part in Color

Ooypunk:
When casting a float to string, PHP respects the locale it is in. This means that the float 0.55 is casted to "0.55" in USA (probably loads more countries, but that is beside the point), but where I live (Netherlands), it becomes "0,55".
This means that, when the input is "55CCCCCC", Color::formatted() outputs 'rgba(204, 204, 204, 0,55)', which in turn is interpreted as 'rgba(204, 204, 204, 0, 55)' (a fifth argument has appeared).

The fix I wrote makes sure that this float is always written as "0.55". Old behaviour is respected: 0.5 becomes "0.5" and 0 becomes "0".

### DIFF
--- a/src/Manipulators/Helpers/Color.php
+++ b/src/Manipulators/Helpers/Color.php
@@ -113,8 +113,8 @@ class Color
     {
         // Make sure the alpha float is correctly formatted
         $formatted = number_format($this->alpha, 2, '.', ',');
-		$zero_trimmed = rtrim($formatted, '0');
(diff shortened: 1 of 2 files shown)


tgalopin:
I'm closing this as it doesn't seem to be a problem to many people (perhaps it was fixed in PHP?). Feel free to open a new PR if you think that's worth it, and thanks for the work!

Issue #223, opened 2018-02-15
cropping issue


juliangums:
when cropping with focal point, it zooms in
![screenshot from 2018-02-15 18-47-45](https://user-images.githubusercontent.com/22907521/36275180-03be3e36-1282-11e8-978f-c2b162770913.png)

however, reverting `intervention/image` to `2.3.4` fixes the issue
![screenshot from 2018-02-15 18-47-45](https://user-images.githubusercontent.com/22907521/36275211-1c8c880a-1282-11e8-8c2b-4bdaa5e4fea2.png)

I understand this is likely to be an issue with `intervention/image`. I don't really know what has changed there but for now I recommend forcing composer to use of `2.3.4` until the issue is fixed on `intervention/image`. Maybe worth creating an issue there. I just don't know how to describe it properly

reinink:
I'm sorry @juliangums for the late reply on this.

The two images you posted look identical to me.

Is this still an issue?

juliangums:
Haha, looks like I accidentally uploaded the same screenshot twice. Basically just imagine the first image being zoomed in cutting off half of the image. This happened when defining a focal point for cropping. I'm not sure if this has been fixed yet in newer versions of `intervention/image`. I'm unable to test it again in the weeks to come.

tgalopin:
Is this issue still present in latest versions of Intervention Image/Glide?

ADmad:
Given that the latest release of `intervention/image` is `2.5.1` I am going to assume that the issue is fixed. Please open a new issue if the problem persists.

PR #224, opened 2018-02-15
Fix cropping issue

juliangums:
described in https://github.com/thephpleague/glide/issues/223


ADmad:
Given that the latest release of `intervention/image` is `2.5.1` I am going to assume that the issue is fixed. Please open a new issue if the problem persists.

PR #228, opened 2018-04-17
Changed manipulators order.

Trigur:
Cropping image with flip:

Expected:

But result is:

So, i changed order of manipulators.

### DIFF
--- a/src/ServerFactory.php
+++ b/src/ServerFactory.php
@@ -218,6 +218,7 @@ class ServerFactory
     public function getManipulators()
     {
         return [
+            new Flip(),
             new Orientation(),
             new Crop(),
             new Size($this->getMaxImageSize()),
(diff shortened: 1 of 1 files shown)


baileylo:
You can specify the order yourself, see the _Manual Setup_ section of the [Setup Instructions](http://glide.thephpleague.com/1.0/config/setup/).

Trigur:
Thanks! But i can't do that. Glide uses by https://github.com/spatie/laravel-medialibrary as composer dependency. And i don't see any ways to specify the order of manipulators.

baileylo:
@Trigur I'd say is an issue with the down stream library, Laravel Media Library. Though I think you can fix it by creating a new media conversion on your model. You'd need to add a `[addMediaConversion](https://github.com/spatie/laravel-medialibrary/blob/master/src/HasMedia/HasMediaTrait.php#L410)` to your model that has the trait. Spatie will know if there's a better way to do this though.

reinink:
Going to close this, since I don't think I want to change this behaviour at this time, and as @baileylo, there are ways to customize this on a per project basis. Thanks anyway!

PR #236, opened 2018-06-26
Ignore notices raised by tempnam

tgalopin:
Since PHP 7.1, tempnam sometimes raise notices to alert when a unexpected behavior is used as a fallback (https://bugs.php.net/bug.php?id=69489).

As this notice isn't much of a problem in Glide (we need the file to be created in the system temporary directory anyway), we can safely ignore these notices.

This notice happens regularly on my Symfony application (and as Symfony catches it in dev, it breaks the page).

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -514,7 +514,7 @@ class Server
         // We need to write the image to the local disk before
         // doing any manipulations. This is because EXIF data
         // can only be read from an actual file.
-        $tmp = tempnam(sys_get_temp_dir(), 'Glide');
+        $tmp = @tempnam(sys_get_temp_dir(), 'Glide');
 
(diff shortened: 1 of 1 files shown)


tgalopin:
May I kindly ping @reinink as you seem to be the merger lately :) ? If you're in vacations/not available, please ignore this ping :) .

baileylo:
I feel like this is a configuration issue I'd rather know about. Your `sys_get_temp_dir` should return a writeable directory for your system. If it doesn't, you should update your `php.ini` to return a directory path used for temporary files.

tgalopin:
Indeed, thanks for the feedback, I had a problem of configuration, my temporary dir was not writable. Sorry for the useless ping :) .

Issue #192, opened 2017-05-04
Glide not removing temp files


ndebarnot:
Hi all,

my /tmp folder gets filled with temporary Glide files such as /tmp/Glide8bQmtG for instance.

How can I find out why they are not removed? (it's not a user rights problem).

Thanks!

wbadrh:
## Deleting cached images

`$server->deleteCache($path);`

Glide does not automatically purge cached images. However, this can be done by your application using the deleteCache() method. Note, grouping cache in folders MUST be enabled in order to delete cached images, which is the default setting.

You can setup a cron job cleaning this folder periodically.

## Disabling the cache

In some situations it may be desirable to disable the cache. For example, you may choose to use a tool like Varnish for caching instead. The best way to do this with Glide is to use an [in-memory adapter](http://flysystem.thephpleague.com/adapter/memory/) for Flysystem. This will prevent any cached images from being saved to your local disk.

`composer require league/flysystem-memory`

```
use League\Flysystem\Filesystem;
use League\Flysystem\Memory\MemoryAdapter;

$server = League\Glide\ServerFactory::create([
    'source' => 'path/to/source/folder',
    'cache' => new Filesystem(new MemoryAdapter()),
]);

$server->outputImage($path, $_GET);
```

http://glide.thephpleague.com/1.0/config/source-and-cache/
http://glide.thephpleague.com/1.0/config/advanced-usage/

ndebarnot:
Hi, thanks for your reply.

The mistake was on my end. Due to wrong file permissions in the cache folder, the resized images could not be moved there and remained in the /tmp folder of the server.

Contrary to what I originally thought it WAS a user rights problem but not in the /tmp folder.

reinink:
@ndebarnot Good to hear you figured it out.

@wbadrh Thanks for stepping in to help out!

gopal-g:
@reinink  May I know what is the idea behind not being able to delete the files with  ```$server->deleteCache($path);``` if group caching being disabled ? How to delete the cached images in such case ?

Art4:
[...]

PR #238, opened 2018-07-25
Always delete tmp file

barryvdh:
If an exception occurs during this try/catch (eg. invalid file type, file too large etc), the /tmp/GlideXXX files will not be deleted. This will fill the /tmp folder and can lead to problems when it's not deleted regularly.

Should fix https://github.com/thephpleague/glide/issues/192

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
   - COMPOSER_OPTS="--prefer-lowest"
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require": {
         "intervention/image": "^2.1",
         "league/flysystem": "^1.0",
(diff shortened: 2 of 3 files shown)


barryvdh:
Oh do we need to support 5.5 still? In that case we should probably catch the exceptions and then rethrow..

barryvdh:
I've updated the config to drop 5.4, but it seems to have errored on Travis. The tests itselfs should run fine now.

barryvdh:
@reinink What do you think about this? Can we drop 5.4 (and 5.5?)

reinink:
Thanks Barrie. Sorry for the delay. 🤦‍♂️

barryvdh:
Barry* ;)
But thanks, better late than never ;)

reinink:
@barryvdh Ahh, sorry Barry! FWIW, I got it right in the [release notes](https://github.com/thephpleague/glide/releases). 😎

PR #251, opened 2018-12-19
Add support for WebP

jbrooksuk:
Replaces https://github.com/thephpleague/glide/pull/202

### DIFF
--- a/src/Manipulators/Encode.php
+++ b/src/Manipulators/Encode.php
@@ -46,6 +46,7 @@ class Encode extends BaseManipulator
             'jpg' => 'image/jpeg',
             'pjpg' => 'image/jpeg',
             'png' => 'image/png',
+            'webp' => 'image/webp',
         ];
 
         if (array_key_exists($this->fm, $allowed)) {
--- a/src/Manipulators/Encode.php
+++ b/src/Manipulators/Encode.php
@@ -46,8 +46,12 @@ class Encode extends BaseManipulator
             'jpg' => 'image/jpeg',
             'pjpg' => 'image/jpeg',
             'png' => 'image/png',
-            'webp' => 'image/webp',
         ];
+
+        // PHP 5.* may not be compiled with WebP support, so check first.
+        if (function_exists('imagecreatefromwebp')) {
+            $allowed['webp'] = 'image/webp';
+        }
 
         if (array_key_exists($this->fm, $allowed)) {
             return $this->fm;
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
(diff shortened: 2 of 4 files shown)


jbrooksuk:
Travis doesn't have WebP support enabled for PHP 5.4 https://github.com/travis-ci/travis-ci/issues/8435

reinink:
> Travis doesn't have WebP support enabled for PHP 5.4

Meaning it's supported in later versions of PHP? Honestly, it's probably getting to the point that we could drop support for PHP 5 entirely.

jbrooksuk:
>> Travis doesn't have WebP support enabled for PHP 5.4 ...

@reinink as in, it PHP 5.4 does support it, but Travis hasn’t built PHP to include the extension required. I’d agree though, 5.* is coming to complete EOL. http://php.net/supported-versions.php

reinink:
Yeah 10-4. It looks like all the PHP 5.X builds are failing:

image

jbrooksuk:
👍 

Thoughts on dropping PHP 5 support in a separate PR?

reinink:
I think I'm going to keep 5.X support in 1.X. I'd like to find a way to make Travis ignore these tests on 5.X somehow.

I know it's hard to believe (), but I'm planning on getting back into my 2.0 release soon, and I think we'll make that a 7.X only version.

src/Manipulators/Encode.php
```diff
@@ -46,6 +46,7 @@ public function getFormat(Image $image)
             'jpg' => 'image/jpeg',
             'pjpg' => 'image/jpeg',
             'png' => 'image/png',
+            'webp' => 'image/webp',
```
jbrooksuk: @reinink we could check for the existence of the [`imagewebp`](http://www.php.net/manual/en/function.imagewebp.php) function before adding this format.

reinink: Yes! I like that. Can you add that?

jbrooksuk: I've added the condition, but not sure if it'll work.

jbrooksuk: I had to add it elsewhere 👍

jbrooksuk:
> I think I'm going to keep 5.X support in 1.X. I'd like to find a way to make Travis ignore these tests on 5.X somehow.

Done!

> I know it's hard to believe (), but I'm planning on getting back into my 2.0 release soon, and I think we'll make that a 7.X only version.

That'd be great. I'd love to be able to add custom commands (like for instance, a soft blur).

reinink:
Tagged: https://github.com/thephpleague/glide/releases/tag/1.4.0

PR #266, opened 2019-09-24
Update .travis.yml

fisharebest:
Three changes to `.travis.yml`:

* Add PHP 7.3 to the build matrix.
* The default environment on TravisCI [changed](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) from `trusty` to `xenial` in April 2019.  This means that PHP 5.5 won't build unless we explicitly choose the older environment.
* Explicitly use our own version of `phpunit`.  Otherwise some builds pick up an incompatible version from the environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,20 @@ env:
   - COMPOSER_OPTS="--prefer-lowest"
 
 php:
-  - 5.5
   - 5.6
   - 7.0
   - 7.1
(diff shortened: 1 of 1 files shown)


tgalopin:
Replaced by migration to GitHub Actions https://github.com/thephpleague/glide/pull/303. But many thanks for the PR!

PR #274, opened 2020-04-16
FIX: resizeCanvas() accepts only integer

racztiborzoltan:
FIX: resizeCanvas() accepts only integer

Sample border parameter that causes the error
```
border=10w,FFCC33,shrink
```

The error: 
resizeCanvas() accepts only integer values as argument 1.

The solution: 
Rounding the value of width in \League\Glide\Manipulators\Border->getBorder() method.

### DIFF
--- a/src/Manipulators/Border.php
+++ b/src/Manipulators/Border.php
@@ -56,7 +56,7 @@ class Border extends BaseManipulator
         $method = $this->getMethod(isset($values[2]) ? $values[2] : null);
 
         if ($width) {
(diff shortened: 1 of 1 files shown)


tgalopin:
I feel like changing it here could have unforeseen consequences on people relying on the Border computation. Could you put the round() call in the resizeCanvas call instead? Thanks!

ADmad:
Fixed in https://github.com/thephpleague/glide/pull/314

Issue #206, opened 2017-09-19
Feature Request: New "stretch-fill" fit method


stevendesu:
Suppose I have an image that is 100 pixels by 300 pixels and I want it to fill a 600 pixel by 600 pixel space.

If the image were the same aspect ratio, but larger, I could use the "fill" fit method to create a 600x600 image with black or white bars on the left and right to fill in where there's no image.

However because my image is *smaller* than the desired final size, "fill" place black/white bars on the top and bottom of the image, as well. This puts the image in a massive white frame.

The only way to actually fill the space is to use "stretch", which loses the aspect ratio and distorts the image.

I would like something that stretches the image only as much as necessary to fill the vertical *or* horizontal space allotted, and fill the remaining area with the background color.

PR #280, opened 2020-07-14
#206: stretch-fill support

xor22h:
Adds support for feature request #206

### DIFF
--- a/src/Manipulators/Size.php
+++ b/src/Manipulators/Size.php
@@ -108,7 +108,7 @@ class Size extends BaseManipulator
      */
     public function getFit()
     {
-        if (in_array($this->fit, ['contain', 'fill', 'max', 'stretch'], true)) {
+        if (in_array($this->fit, ['contain', 'fill', 'max', 'stretch', 'stretch-fill'], true)) {
             return $this->fit;
         }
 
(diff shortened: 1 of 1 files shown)


reinink:
Hey thanks for putting this together! I'll need to do some testing before merging it in.

I'm also not sure I'm sold on the name "stretch-fill", since it sounds like the image is not being stretched (distorted), but rather just scaled, yes?

It would also be great to have a test to go with this.

xor22h:
@reinink yeah, i'm not sure about the title too. It works more like contain-fill :) It was just the behaviour and name suggested by #206;

Because now it works same as contain, just contain, makes image smaller (crops transparent sides), and this one just fills it with a background color making image being the same size as requested.

reinink:
Understood, okay, let me think on this.

Fair warning, I'm busy this week, and on holidays next week, so it will likely be the week after. 👍

xor22h:
I don't worry too much. I installed it from my own fork so i'm happy for myself :D just saw a related ticket :)

tgalopin:
@reinink did you hve time to check this? I can have a look and merge if you're OK with it :)

reinink:
@tgalopin Negative, never had a chance. Feel free to proceed.

Note, maybe of the terminology that I've used for Glide is based on Imgix's API. I found they've done a pretty good job with their API. Obviously it doesn't have to match exactly, but it might be good to have a peak at their docs for some inspiration here. 👍

tgalopin:
Thanks for the tip, I'll have a look!

ADmad:
Implemented with name "fill-max" in #332.

Issue #170, opened 2017-01-16
Brainstorming on Glide 2.0


reinink:
I've started thinking about Glide 2.0. The purpose of this issue is to share those ideas and get some discussion going about it.

### Serve cached images via web server

My number one goal for Glide 2.0 is allowing cached images to be served by the web server, instead of Glide. Meaning the first request to a manipulated image would hit PHP (and Glide), but all subsequent requests will be handled by the web server automatically. Further, beyond the standard front-controller rewrite rule (routing all requests through index.php), I don't want there to be _any_ extra web server configuration required.

The simplest way to accomplish this is to save the cached images in the same place, and with the exact same URL as the requested images. This way, when the first request comes in, the cached image won't exist, so the request will be routed to PHP and handled by Glide. However, all subsequent requests will hit the cached image directly, without hitting PHP. This should greatly improve the performance of loading cached images.

### New URL format

This will not be possible with `GET` parameters since they are not treated as part of the filename. Further, the proper file format extension (jpg, gif, png) is also required so that the web server will output it properly. Finally, this will also require saving images in a public folder. Here is a proposed format that I've been testing with that works well:

```
http://example.com/img/kayaks.jpg/w-800-h-500-fit-crop-s-5ac7f6163057bf1931fd5b513f797644939ece7724ee3846a27e9915f3ea5eda.jpg
```

It's obviously not as nice as the current GET based URLs, but maybe that's okay. By grouping the cached images into source image folders it's easy to delete all the cached images for a specific image.

### URL generation

I really feel like URL generation using a helper method is going to solve a lot of issues. Right now Glide ships with a URL generator, but it's sort of buried in the [security documentation](http://glide.thephpleague.com/1.0/config/security/). I think this has been a bit of an issue with Glide 1.0. It makes it look like you can easily just edit URLs, but in reality if you sign your images (and you absolutely should be) you have to use some type of URL generator.

By making a URL generation helper more of a first-class feature, it will help minimize the pain caused by the less desirable URL format. Plus it will make HTTP signatures easier to work with. It will look something like this:

```php
<img src="<?=glide('landscapes/lake.jpg', ['w' => 800])?>">
```
```blade
<img src="{{ glide('landscapes/lake.jpg', ['w' => 800]) }}">
```

### HTTP signature simplification

[...]

PR #282, opened 2020-07-21
Custom path callable

mlazze:
This would allow to customize the cache path fully.

An alternative would be to call the callable at the end of getCachePath with all the calculated parts (sourcePath, md5, groupCacheInFolders, cachePathPrefix, cacheWithFileExtensions, ext).

I also bound this to the Server instance so that we may its method within the callable.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -376,10 +376,10 @@ class Server
     public function getCachePath($path, array $params = [])
     {
         $customCallable = $this->getCachePathCallable();
-        if ($customCallable instanceof \Closure) {
+        if (null !== $customCallable) {
             $boundCallable = \Closure::bind($customCallable, $this, self::class);
 
-            return $boundCallable(...func_get_args());
+            return $boundCallable($path, $params);
         }
         $sourcePath = $this->getSourcePath($path);
 
--- a/src/Server.php
+++ b/src/Server.php
@@ -98,7 +98,7 @@ class Server
     /**
      * Custom cache path callable.
      *
-     * @var callable|null
+     * @var \Closure|null
      */
     protected $cachePathCallable;
 
@@ -348,9 +348,9 @@ class Server
     /**
      * Set a custom cachePathCallable.
      *
-     * @param callable|null $cachePathCallable The custom cache path callable. It receives the same arguments as @see getCachePath
+     * @param \Closure|null $cachePathCallable The custom cache path callable. It receives the same arguments as @see getCachePath
      */
-    public function setCachePathCallable($cachePathCallable)
+    public function setCachePathCallable(?\Closure $cachePathCallable)
     {
         $this->cachePathCallable = $cachePathCallable;
     }
@@ -358,7 +358,7 @@ class Server
     /**
      * Gets the custom cachePathCallable.
      *
-     * @return callable|null The custom cache path callable. It receives the same arguments as @see getCachePath
+     * @return \Closure|null The custom cache path callable. It receives the same arguments as @see getCachePath
      */
     public function getCachePathCallable()
     {
@@ -376,7 +376,7 @@ class Server
     public function getCachePath($path, array $params = [])
     {
         $customCallable = $this->getCachePathCallable();
-        if (is_callable($customCallable)) {
+        if ($customCallable instanceof \Closure) {
             $boundCallable = \Closure::bind($customCallable, $this, self::class);
 
             return $boundCallable(...func_get_args());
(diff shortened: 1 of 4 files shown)


tgalopin:
What's your use case for this? I'm not fundamentally against adding it, but I don't think users need should care about the internal structure of the cache. I may be missing a valid use case though.

ADmad:
@tgalopin One of goals for Glide 2.0 stated by @reinink in #170 is: `Serve cached images via web server`. 

Until the same is supported/implemented, allowing users to customize the cache path using a callable could help users achieve the same in the meanwhile.

tgalopin:
That's an interesting idea indeed. @mlazze are you still interested in finalizing this PR? It needs rebasing and I have some comments on the code but I prefer to ask before moving forward on this PR :) .

mlazze:
Yes, i'm customizing the path callable to allow the ws to cache images.

I'm willing to finalize this PR and rebase it

mlazze:
Rebased on master.

7.2 fails for missing gd library, i'm not sure why

src/Server.php
```diff
@@ -319,6 +325,26 @@ public function getCacheWithFileExtensions()
         return $this->cacheWithFileExtensions;
     }
 
+    /**
+     * Set a custom cachePathCallable.
+     *
+     * @param callable|null $cachePathCallable The custom cache path callable. It receives the same arguments as @see getCachePath
```
ADmad, line 331: I would prefer restricting this to `Closure` instead of `callable` and add `Closure` type for the argument in method signature. `callable` allows a bazillions things making `callable` type checking expensive. 

Users can always use `Closure::fromCallable()` for the method call if they really need to use other callable types.

ADmad:
> 7.2 fails for missing gd library, i'm not sure why

Rebasing your branch onto upstream master should fix that.

mlazze:
Hello, just switched from callable to Closure and rebased on master.

src/Server.php
```diff
@@ -349,6 +375,12 @@ public function getCacheWithFileExtensions()
      */
     public function getCachePath($path, array $params = [])
     {
+        $customCallable = $this->getCachePathCallable();
+        if ($customCallable instanceof \Closure) {
```
ADmad, line 379:
```suggestion
        if (null !== $customCallable) {
```

src/Server.php
```diff
@@ -349,6 +375,12 @@ public function getCacheWithFileExtensions()
...
+            $boundCallable = \Closure::bind($customCallable, $this, self::class);
+
+            return $boundCallable(...func_get_args());
```
ADmad, line 382: Why use `func_get_args()` when the 2 arguments `$path` and `$params` are already available?

src/ServerFactory.php
```diff
@@ -149,6 +150,18 @@ public function getTempDir()
...
     }
 
+    /**
+     * Get cache path callable.
+     *
+     * @return \Closure|null Cache path callable.
+     */
+    public function getCachePathCallable()
+    {
+        if (isset($this->config['cache_path_callable'])) {
```
ADmad, line 160: A single line will suffice `return $this->config['cache_path_callable'] ?? null;`

src/Server.php
mlazze: Mostly just to make sure we're passing all the arguments in the same order as the main function, even if more get added.
I can change that, if it doesn't suit the needs.

ADmad: It's just 2 params, let's avoid an unneeded function call :slightly_smiling_face:.

mlazze: :+1:

mlazze:
I may also add return type hints if needed, but i'm not seeing them in the codebase.

src/Server.php
```diff
@@ -349,6 +375,12 @@ public function getCacheWithFileExtensions()
      */
     public function getCachePath($path, array $params = [])
     {
+        $customCallable = $this->getCachePathCallable();
+        if (null !== $customCallable) {
+            $boundCallable = \Closure::bind($customCallable, $this, self::class);
```
ADmad, line 380: Since you are binding current object's context can you please add a test for it too. For e.g. calling `$this->getSourcePath()` inside the callable.

ADmad: @mlazze Ping

huiyang:
Is there any update? 
I need this update to serve image cache directly from web server without go through php.
Is the review pending changes from @mlazze ?

Appreciate for the effort anyway.

huiyang:
Is there any update for this?

Thanks.

PR #283, opened 2020-07-21
read the image directly from S3, don't write to a tmpfile

mattbucci:
tmpfile seems unnecessary GD can use the s3 stream directly with no issues.


mattbucci:
meant to open PR against our fork, please ignore

PR #300, opened 2021-01-19
Flysystem v2

Lustmored:
This is alternative approach to Flysystem v2 integration based on discussion in #291.

For now it's just a draft as none of the tests are refactored to handle new interfaces. Before touching them I'd love to hear opinions about code changes. I had no idea for better `ResponseFactoryInterface`, open to suggestions.

One important thing I might have missed some place is exception handling - Flysystem v2 changed a lot and now throws exceptions instead of returning `false`. I've tried to adapt to it, but probably missed a place or two.

And question about tests - how to proceed? Shall I detect available Flysystem version and check against available interfaces or write new tests for Flysystem v2 and skip those unavailable? Or maybe some other way? I think checking available interfaces is the way to go as we can then set up CI jobs to test against different Flysystem versions, but I might be wrong.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -163,7 +163,7 @@ class Server
      */
     public function sourceFileExists($path)
     {
-        if($this->source instanceof FilesystemInterface) {
+        if ($this->source instanceof FilesystemInterface) {
             return $this->source->has($this->getSourcePath($path));
         }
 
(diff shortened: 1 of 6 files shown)


src/Server.php
```diff
@@ -184,16 +190,16 @@ public function getBaseUrl()
 
     /**
      * Set cache file system.
-     * @param FilesystemInterface $cache Cache file system.
+     * @param FilesystemInterface|FilesystemOperator $cache Cache file system.
      */
-    public function setCache(FilesystemInterface $cache)
+    public function setCache($cache)
```
ADmad, line 195: While the chances of some extending the `Server` class are quite low, changing the signature of a public method is a BC break as per semver.

tgalopin:
Closing as as discussed, we'll create a release focused on Flysystem 2 support (https://github.com/thephpleague/glide/pull/301)

PR #301, opened 2021-01-20
Use only Flysystem v2

Lustmored:
This PR contains changes that remove Flysystem 1 and adapt to Flysystem 2.

Most of the work went into changing test suite to mimic Flysystem 2 exception logic, while changes to the code are rather simple.

I create another PR to visualize changes needed to migrate fully from V1 to V2 in case this will be the way chosen by maintainers.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -3,8 +3,8 @@
 namespace League\Glide;
 
 use InvalidArgumentException;
-use League\Flysystem\FileExistsException;
-use League\Flysystem\FilesystemInterface;
+use League\Flysystem\FilesystemException as FilesystemV2Exception;
+use League\Flysystem\FilesystemOperator;
 use League\Glide\Api\ApiInterface;
 use League\Glide\Filesystem\FileNotFoundException;
 use League\Glide\Filesystem\FilesystemException;
@@ -15,7 +15,7 @@ class Server
     /**
      * Source file system.
      *
-     * @var FilesystemInterface
+     * @var FilesystemOperator
      */
     protected $source;
 
@@ -29,7 +29,7 @@ class Server
     /**
      * Cache file system.
      *
-     * @var FilesystemInterface
+     * @var FilesystemOperator
      */
     protected $cache;
 
@@ -99,11 +99,11 @@ class Server
     /**
      * Create Server instance.
      *
-     * @param FilesystemInterface $source Source file system.
-     * @param FilesystemInterface $cache  Cache file system.
+     * @param FilesystemOperator  $source Source file system.
+     * @param FilesystemOperator  $cache  Cache file system.
      * @param ApiInterface        $api    Image manipulation API.
      */
-    public function __construct(FilesystemInterface $source, FilesystemInterface $cache, ApiInterface $api)
+    public function __construct(FilesystemOperator $source, FilesystemOperator $cache, ApiInterface $api)
     {
         $this->setSource($source);
         $this->setCache($cache);
@@ -114,9 +114,9 @@ class Server
     /**
      * Set source file system.
      *
-     * @param FilesystemInterface $source Source file system.
+     * @param FilesystemOperator $source Source file system.
      */
-    public function setSource(FilesystemInterface $source)
+    public function setSource(FilesystemOperator $source)
     {
         $this->source = $source;
     }
@@ -124,7 +124,7 @@ class Server
     /**
      * Get source file system.
      *
-     * @return FilesystemInterface Source file system.
+     * @return FilesystemOperator Source file system.
      */
     public function getSource()
     {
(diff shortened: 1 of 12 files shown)


src/Server.php
```diff
@@ -523,20 +543,14 @@ public function makeImage($path, array $params)
...
-            if ($write === false) {
-                throw new FilesystemException(
-                    'Could not write the image `'.$cachedPath.'`.'
-                );
-            }
-        } catch (FileExistsException $exception) {
-            // This edge case occurs when the target already exists
-            // because it's currently be written to disk in another
-            // request. It's best to just fail silently.
+        } catch (FilesystemV2Exception $exception) {
```
tgalopin, line 550: Is it normal that we don't handle the edge case anymore?

tgalopin:
This is a great PR, thanks a lot for your work (again!). I feel the option of releasing Glide 2.0 with only Flysystem 2 is probably the best idea, and then use 3.0 as an opportunity to add other BC breaks if necessary.

WDYT?

src/Server.php
Lustmored: In Flysystem 2 there is no distinction between writing and updating files. Flysystem 1 would fail to write to existing file with FileExistsException, while in Flysystem 2 `write()` works like `put()` in first version - it either creates or updates file. Therefore FileExistsException does not exists anymore and the process can only throw if writing failed, so edge case probably doesn't exists.

But if we want to avoid writing over existing files I can wrap `write` call inside `!fileExists` if clause - I think it can be beneficial.

Lustmored:
> This is a great PR, thanks a lot for your work (again!). I feel the option of releasing Glide 2.0 with only Flysystem 2 is probably the best idea, and then use 3.0 as an opportunity to add other BC breaks if necessary.
> 
> WDYT?

I agree. While other PR shows that glide can support both versions, it'd probably be a burden to maintain.

tgalopin:
FYI I'm merging several PRs in Glide 1.0, trying to make a final release that provides features for the ones who asked before releasing Glide 2, as it would require them to migrate to Flysystem 2.

Can you finalize this PR if you didn't already and rebase when you have time? I'll merge as soon as possible :) .

Lustmored:
I have rebased to master and finalized PR as you requested. I have opted to not make any change regarding your comment as that'd require changes in tests that I am uncertain about.

Lustmored:
@tgalopin just a friendly ping :)

tgalopin:
Thanks! I'll merge this asap :) . Pinning the tab on my browser ;) .

src/Server.php
```diff
@@ -3,8 +3,8 @@
 namespace League\Glide;
 
 use InvalidArgumentException;
-use League\Flysystem\FileExistsException;
-use League\Flysystem\FilesystemInterface;
+use League\Flysystem\FilesystemException as FilesystemV2Exception;
```
rjd22, line 6: Maybe it would be more clear to rename this to or `FlysystemException` or `FlysystemFilesystemException` since V2 doesn't have any meaning in Glide.

rjd22 (commented):
@tgalopin @Lustmored I have a small need for this. So I also gone over the changes but they look good to me. I've added a really small nitpick.

Would it be possible to merge this?

src/Server.php
Lustmored: I don't have strong preference here, yet it's just an alias and aliases are scoped per file. Thus consumers of the library still will have to catch `League\Flysystem\FilesystemException` and the alias `FilesystemV2Exception` won't be exposed anywhere. To me it was useful name as it states exactly where the exception comes from.

But still I don't have strong preference :)

rjd22: It was a nitpick so LGTM.

rjd22:
@tgalopin I don't know hounding maintainers because I know we're all busy. But can I help in some way with checking and merging pull-requests? I'm actively using this library.

ADmad:
It would be great if there's more than one individual who has the rights to merge at least maintenance related pull request. Given the popularity of the lib and the limited amount of time the past and current maintainer has had, it takes a bit too long for PRs to be reviewed/merged. PRs such as this one in particular would be blocking development for so many projects.

tgalopin:
I agree, I don't have much time to maintain this library in the short term. @ADmad would you be down for a quick chat in the coming days/weeks to discuss this? Can you ping me at galopintitouan@gmail.com ?

tgalopin:
Thanks @Lustmored !

rjd22:
@tgalopin I would also not mind spending time to review, and merge PR's. If you're willing to have me.

tgalopin:
@rjd22 thanks a lot for your proposal, that's very much appreciated!

I think we need to be careful not to add too many people in the team, as it can create more problems than it would solve, and in this regard the overall big involvement of @ADmad in the project led me to think we should discuss it together first. If we think we need a third person, I'll definitely ping you :) !

PR #307, opened 2021-02-02
PHP 8 Support for Glide v2

nagibmahfuj:
Check this to accept merge request for v2


tgalopin:
Hi @nagibmahfuj !

2.0 isn't released yet and the current branch for 2.0 is going to be deleted at some point as we will release Glide 2.0 as the same library as Glide 1 but with Flysystem 2.0 support. But thanks for the PR!

PR #310, opened 2021-02-17
Fix rounding issue with Manipulators\Size

sebastianpopp:
If I have an image with a size of 1024x553 pixel and I run

```php
(new \League\Glide\Manipulators\Size)
    ->setParams(['w' => 411, 'fit' => 'max'])
    ->run($image);
```

on it. I will get an image with the dimensions 409x221 pixel.

It looks like this happens due to the calculations in \League\Glide\Manipulators\Size::resolveMissingDimensions(). For the example above it will keep the width of 411 px and calculate a height of 221 px. Intervention Image rounds numbers differently, resulting in this wrong image size when the image is actually resized to the precalculated size.

To fix this issue I would suggest to rely on the resizing logic of Intervention Image.

### DIFF
--- a/tests/Manipulators/SizeTest.php
+++ b/tests/Manipulators/SizeTest.php
@@ -126,6 +126,16 @@ class SizeTest extends TestCase
         $this->assertSame([200, 100], $this->manipulator->resolveMissingDimensions($image, null, 100));
     }
 
+    public function testResolveMissingDimensionsWithOddDimensions()
+    {
+        $image = Mockery::mock('Intervention\Image\Image', function ($mock) {
+            $mock->shouldReceive('width')->andReturn(1024);
+            $mock->shouldReceive('height')->andReturn(553);
+        });
+
+        $this->assertSame([411, 222], $this->manipulator->resolveMissingDimensions($image, 411, null));
+    }
+
     public function testLimitImageSize()
     {
         $this->assertSame([1000, 1000], $this->manipulator->limitImageSize(1000, 1000));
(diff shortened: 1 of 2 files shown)


tests/Manipulators/SizeTest.php
```diff
@@ -126,6 +126,16 @@ public function testResolveMissingDimensions()
...
 
+    public function testResolveMissingDimensionsWithOddDimensions()
+    {
+        $image = Mockery::mock('Intervention\Image\Image', function ($mock) {
+            $mock->shouldReceive('width')->andReturn(1024);
+            $mock->shouldReceive('height')->andReturn(553);
+        });
+
+        $this->assertSame([411, 222], $this->manipulator->resolveMissingDimensions($image, 411, null));
+    }
```
rjd22, line 137: @sebastianpopp thx the fix. Might it be a good idea to also add tests for a resize with a null width?

rjd22:
I saw the GD issue was fixed so if you rebase it on the last master most likely the CI will pass.

tests/Manipulators/SizeTest.php
sebastianpopp: There's already a test for a resize with a null width above. I just wanted to add an additional test case that fails with the old implementation and passes with my fix to demonstrate why it's needed.

tgalopin (approved):

ADmad (approved):

ADmad:
Thanks for nudging this along @rjd22 :slightly_smiling_face:.

PR #312, opened 2021-03-03
Add static analysis using Psalm

ADmad:
There are quite a few `int` vs `float` type issues reported at higher level of psalm but I will make a follow up PR for that for easier reviewing.

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -38,6 +38,8 @@ class Api implements ApiInterface
      * Set the image manager.
      *
      * @param ImageManager $imageManager Intervention image manager.
+     *
+     * @return void
      */
     public function setImageManager(ImageManager $imageManager)
     {
@@ -58,6 +60,8 @@ class Api implements ApiInterface
      * Set the manipulators.
      *
      * @param array $manipulators Collection of manipulators.
+     *
+     * @return void
      */
     public function setManipulators(array $manipulators)
     {
--- a/src/Responses/PsrResponseFactory.php
+++ b/src/Responses/PsrResponseFactory.php
@@ -4,7 +4,6 @@ namespace League\Glide\Responses;
 
 use Closure;
 use League\Flysystem\FilesystemOperator;
-use League\Glide\Filesystem\FilesystemException;
 use Psr\Http\Message\ResponseInterface;
 
 class PsrResponseFactory implements ResponseFactoryInterface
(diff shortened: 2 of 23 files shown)


src/Api/Api.php
```diff
@@ -58,6 +60,8 @@ public function getImageManager()
      * Set the manipulators.
      *
      * @param array $manipulators Collection of manipulators.
+     *
+     * @return void
```
rjd22, line 64: Can't we start using real return types now, since the minimal PHP requirement is 7.2?

ADmad: Personally I would love to add real param and return types :) Hopefully we can do so soon.

src/Responses/PsrResponseFactory.php
```diff
@@ -54,14 +53,6 @@ public function create(FilesystemOperator $cache, $path)
         $cacheControl = 'max-age=31536000, public';
         $expires = date_create('+1 years')->format('D, d M Y H:i:s').' GMT';
 
-        if (false === $contentType) {
-            throw new FilesystemException('Unable to determine the image content type.');
-        }
-
-        if (false === $contentLength) {
-            throw new FilesystemException('Unable to determine the image content length.');
-        }
```
rjd22, line 63: I hope you don't mind me asking but doesn't removing this cause any issues? Because file size van still be `null`?

https://github.com/thephpleague/flysystem/blob/7cbbb7222e8d8a34e71273d243dfcf383ed6779f/src/FileAttributes.php#L72-L75

Someone might be catching this.

ADmad: `FileAttributes` is an internal class. `Filesystem::fileSize()` has it's return value typed as `int` https://github.com/thephpleague/flysystem/blob/7cbbb7222e8d8a34e71273d243dfcf383ed6779f/src/Filesystem.php#L118. So it won't return `null`.

ADmad: There's a possibility that the adapter could throw an exception if it's unable to access the file, so we could wrap those calls in try/catch and throw own own `League\Glide\Filesystem\FilesystemException` exception if we want to.

src/Api/Api.php
tgalopin: Let's add psalm in the minor and types in the next major IMO

src/Responses/PsrResponseFactory.php
tgalopin: I don't think try catching here is necessary: this probably changed in Flysystem 2.0 anyway, if people upgrade to Glide 2.0 they'll need to adapt their code here, that's fine IMO.

tgalopin (approved):

PR #318, opened 2021-03-29
Update league/flysystem version

mwangaben:


mwangaben:
nice

ADmad:
Why is this required?

tgalopin:
I don't think it is: the test suite passes on the lowest deps. I'm closing this PR but @mwangaben if you have more details feel free to tell us more.

PR #325, opened 2021-06-10
Allow mount identifier, for MountManager of FlySystem library, into prefixes

ElGigi:
Allow mount identifier, for MountManager of FlySystem library, into prefixes.

```php
$server->setSourcePathPrefix('source://');
$server->setCachePathPrefix('cache://');
```


src/Server.php
```diff
@@ -111,6 +111,17 @@ public function __construct(FilesystemOperator $source, FilesystemOperator $cach
         $this->tempDir = sys_get_temp_dir();
     }
 
+    private function trimPrefix(string $prefix): string
+    {
+        if ('//' == substr($prefix, -2)) {
```
ADmad, line 116: Why not check for `'://' === substr($prefix, -3)` instead of having a separate check for `:`?

ElGigi:
A little push to allow mount identifier into path without set prefix.
Example: get file "img://my/img.jpg"

If i set a cache prefix ; the mount identifier is removed.

Example: set cache prefix "cache://my/dir" and get "img://my/img.jpg" , the cache path will be "cache://my/dir/my/img/jpg"

ElGigi:
Any news? Thx

src/Server.php
ADmad: Can you respond to above query please.

ElGigi:
@ADmad Is there a chance that this proposal will be accepted?

src/Server.php
```diff
@@ -357,12 +381,16 @@ public function getCachePath(string $path, array $params = []): string
 
         $cachedPath = hash('xxh3', $sourcePath.'?'.http_build_query($params));
 
+        if (false !== strpos($sourcePath, '://')) {
```
ADmad: A micro optimization: `strpos()` already gives you the position, so use that with `substr()` to genereate `$sourcePath` instead of using `explode()`.

Similar can be done in `Server::removeFilesystemIdentifier()`

ADmad:
> Is there a chance that this proposal will be accepted?

Thank you for your patience. It can be merged once the minor nitpick is addressed.

ElGigi:
Thanks! Done 👍

PR #332, opened 2021-10-07
Add "fill-max" resizing option

kjellknapen:
I noticed the issue for #206 had a pull request but it hadn't been merged yet because of naming issues. So I went to check in the Imgix api and made changes accordingly. In the imgix api the fill function is used for resizing and filling an image (so this function shoud actually do what is requested in the issue). 

So in my PR I changed this function to the function made by @xor22h in the previous PR. Next to that I added the fillmax function to allow users to also fill and preserve ratio as the fill function used to do. If you rather have them the other way around I can change it since this would be a change affecting everyone using the fill method.


kjellknapen:
@ADmad You are right, I switched them up so it won't result in a breaking change :)

ADmad:
Wonder why PHP 8.1 is segfaulting. Also can you update the docs too regarding the new option.

docs/1.0/api/size.md
```diff
@@ -34,6 +34,7 @@ Sets how the image is fitted to its target dimensions.
 - `contain`: Default. Resizes the image to fit within the width and height boundaries without cropping, distorting or altering the aspect ratio.
 - `max`: Resizes the image to fit within the width and height boundaries without cropping, distorting or altering the aspect ratio, and will also not increase the size of the image if it is smaller than the output size. 
 - `fill`: Resizes the image to fit within the width and height boundaries without cropping or distorting the image, and the remaining space is filled with the background color. The resulting image will match the constraining dimensions.
+- `fill-max`: Resizes the image to fit within the width and height boundaries with cropping but without distorting the image. The resulting image will match the constraining dimensions.
```
ADmad, line 37: This page is for v1. The feature is not going to be backported.

docs/2.0/api/size.md
```diff
@@ -34,6 +34,7 @@ Sets how the image is fitted to its target dimensions.
 - `contain`: Default. Resizes the image to fit within the width and height boundaries without cropping, distorting or altering the aspect ratio.
 - `max`: Resizes the image to fit within the width and height boundaries without cropping, distorting or altering the aspect ratio, and will also not increase the size of the image if it is smaller than the output size. 
 - `fill`: Resizes the image to fit within the width and height boundaries without cropping or distorting the image, and the remaining space is filled with the background color. The resulting image will match the constraining dimensions.
+- `fill-max`: Resizes the image to fit within the width and height boundaries with cropping but without distorting the image. The resulting image will match the constraining dimensions.
```
ADmad, line 37: The code checks for `fillmax` without the `-`. So the docs and code need to be harmonized. 

Personally I would prefer using `fill-max` since other multi word values for other options are also dash separated, for e.g. `top-left`.

src/Manipulators/Size.php
```diff
@@ -309,6 +313,24 @@ public function runFillResize($image, $width, $height)
...
+    /**
+     * Perform fillmax resize image manipulation.
+     *
+     * @param Image $image  The source image.
+     * @param int   $width  The width.
+     * @param int   $height The height.
+     *
+     * @return Image The manipulated image.
+     */
+    public function runFillMaxResize($image, $width, $height)
```
ADmad, line 325:
```suggestion
    public function runFillMaxResize(Image $image, $width, $height)
```

docs/1.0/api/size.md
kjellknapen: You are right, I have removed it

docs/2.0/api/size.md
kjellknapen: I have updated all other references to fill-max

ADmad: > .. but without distorting the image

This seems a bit inaccurate. With `fill-max` smaller images will be stretched isn't it (while still maintaining aspect ratio)? So technically that will distort/blur the image.

kjellknapen: Yeah it's inaccurate. I should change it to without cropping but with distorting. And add that the remaining width or height will be filled with a color

docs/2.0/api/size.md
```diff
@@ -34,6 +34,7 @@ Sets how the image is fitted to its target dimensions.
 - `contain`: Default. Resizes the image to fit within the width and height boundaries without cropping, distorting or altering the aspect ratio.
 - `max`: Resizes the image to fit within the width and height boundaries without cropping, distorting or altering the aspect ratio, and will also not increase the size of the image if it is smaller than the output size. 
 - `fill`: Resizes the image to fit within the width and height boundaries without cropping or distorting the image, and the remaining space is filled with the background color. The resulting image will match the constraining dimensions.
+- `fill-max`: Resizes the image to fit within the width and height boundaries without cropping but with distorting/blurring the image if it's smaller. The finished image will have remaining space on either width or height (except if the aspect ratio of the new image is the same as the old image). The remaining space wil be filled with the background color. The resulting image will match the constraining dimensions.
```
ADmad, line 37:
```suggestion
- `fill-max`: Resizes the image to fit within the width and height boundaries without cropping but with distorting/blurring the image if it's smaller. The finished image will have remaining space on either width or height (except if the aspect ratio of the new image is the same as the old image). The remaining space will be filled with the background color. The resulting image will match the constraining dimensions.
```

ADmad:
Please rebase your branch onto upstream master where I have fixed the testsuite failure on PHP 8.1.

kjellknapen:
> Please rebase your branch onto upstream master where I have fixed the testsuite failure on PHP 8.1.

Rebased everything!

ADmad:
Sorry I didn't notice this earlier but codecov says there's no test coverage for the new `runFillMaxResize()` method. Please add a test for it.

kjellknapen:
@ADmad Is it okay now?

ADmad:
Looks like an invocation count needs to be updated to fix the test failure.

ADmad:
@kjellknapen This will go faster if you run the test suite locally first to ensure it passes :slightly_smiling_face:.

kjellknapen:
@ADmad Yeah I know, I'm sorry, thought it was going to be a quick and easy fix! Now it should work. Sorry that everything took so long, haven't done a whole lot of open source!

ADmad:
@kjellknapen No need to apologize, thank you for seeing this through.

PR #337, opened 2021-12-09
Fix php8.1 deprecations

bastien-phi:
Fixes deprecations for php8.1

### DIFF
--- a/src/Manipulators/Brightness.php
+++ b/src/Manipulators/Brightness.php
@@ -34,11 +34,7 @@ class Brightness extends BaseManipulator
      */
     public function getBrightness()
     {
-        if (null === $this->bri) {
-            return;
-        }
-
-        if (!preg_match('/^-*[0-9]+$/', $this->bri)) {
+        if (null === $this->bri || !preg_match('/^-*[0-9]+$/', $this->bri)) {
             return;
         }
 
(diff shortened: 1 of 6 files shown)


src/Manipulators/Brightness.php
```diff
@@ -34,6 +34,10 @@ public function run(Image $image)
      */
     public function getBrightness()
     {
+        if (null === $this->bri) {
```
ADmad, line 37: A separate `if` block isn't necessary. You can just add `null === $this->bri || ` to the `if` block below.

ADmad:
Please rebase from upstream master where I have fixed the testsuite failure on PHP 8.1.

bastien-phi:
Should be all good !

src/Manipulators/Gamma.php
```diff
@@ -34,6 +34,10 @@ public function run(Image $image)
      */
     public function getGamma()
     {
+        if (null === $this->gam) {
+            return;
+        }
+
         if (!preg_match('/^[0-9]\.*[0-9]*$/', $this->gam)) {
```
ADmad, line 41: ... and here.

src/Manipulators/Contrast.php
```diff
@@ -34,6 +34,10 @@ public function run(Image $image)
      */
     public function getContrast()
     {
+        if (null === $this->con) {
+            return;
+        }
+
         if (!preg_match('/^-*[0-9]+$/', $this->con)) {
```
ADmad, line 41: Same change as the one I suggested for `Brightness` here too :slightly_smiling_face:.

ADmad:
No problem, thanks.

bastien-phi:
Glad to help !

PR #345, opened 2022-02-20
Enable TIFF image format handling, fixes #344

### LINKED ISSUE #344: Saving TIFF images
I'm using Glide through spatie/image, and I need to be able to save TIFF format images. Having found #149 which said that all I needed to do was use the imagick driver instead of gd (because imagick supports TIFF and gd does not), I switched to imagick. I was then a bit mystified to find that while it produced files with a `.tiff` extension, they actually contain JPEG data.

I tracked this down to [this code](https://github.com/thephpleague/glide/blob/master/src/Manipulators/Encode.php#L48), which restricts the available output formats, but also makes the output default to JPEG data without changing the filename, which explains my symptoms. Adding TIFF to that array allows it to work, but I can't see how it could work without it, making the answer to that issue a mystery!

PR incoming...

Synchro:
Further to #344, here's a PR to enable TIFF image format support. I saw that there were some vestiges of TIFF format support already (for example the `tif` property in the test case), so I've made use of that. The test suite runs only with the `gd` driver, so operations that require `imagick` (which includes the TIFF format) can't be tested, so I wrote a new test method that reconfigures the test to run with imagick, and skips the test if it's not installed.

I don't understand how the mock works in `testGetFormat` – I tried copying it but couldn't make it work, so I've not created a similar test for that.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,7 +14,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '7.4'
-                  extensions: gd imagick
+                  extensions: gd, imagick
             - name: php-cs-fixer
               run: |
                   wget https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2.16.4/php-cs-fixer.phar -q
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,7 +14,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '7.4'
-                  extensions: gd
+                  extensions: gd imagick
             - name: php-cs-fixer
               run: |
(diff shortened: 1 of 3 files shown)


ADmad:
Please add imagick to the extensions list here to ensure it's available https://github.com/thephpleague/glide/blob/3aa453969f28ae3b9f3a1fdcfca0147ee5454567/.github/workflows/test.yaml#L39

Synchro:
I've added that.

.github/workflows/test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '7.4'
-                  extensions: gd
+                  extensions: gd imagick
```
ADmad, line 17: It's a CSV list https://github.com/shivammathur/setup-php#heavy_plus_sign-php-extension-support

Synchro: Duh, sorry. I corrected that, and added it in the other places it's set too.

Synchro:
Thanks

ADmad:
Thank you! New release done.

Issue #328, opened 2021-07-17
Heroku preview app is dead


bradjones1:
As referenced in docs: https://glide.herokuapp.com

jackmcdade:
Came here to mention this too.

ADmad:
@tgalopin @reinink Can you guys look into this please?

reinink:
So, it looks to me like the docs website was updated to set `/2.0/` in the preview links, but that app was never updated to support version 2. One option here would be to update all those links back to `/1.0` (at least where possible). The better long term option is to update the app that runs on Heroku. To date, that demo app repo has only existed on my system. Should I create a new repo within the @thephpleague org for this?

blite:
still broken

joeworkman:
@reinink I submitted a simple PR to revert the demo images to 1.0 API. It's at least a fix until you can update the app.

ADmad:
Now that heroku killed free accounts the docs site is gone altogether.

wesort:
Still an issue with images not working on docs site.

PR #355, opened 2022-10-03
Revert images in demo to 1.0 API

joeworkman:
This is only until 2.0 can be added to the demo app. 

Related to #328


joeworkman:
Thanks @ADmad

PR #378, opened 2024-01-15
Bump intervention/image version to 3.2

riptin:


ADmad:
Simply bumping the `intervtion/image` isn't going to be enough. Lot of code change will also be required to accommodate the API changes done in the new major version of `intervtion/image`.

Art4:
@ADmad `intervention/image` requires PHP ^8.1, so glide has to drop support for PHP 7.4 and 8.0 too. Are there any objections?

ADmad:
> Are there any objections?

When can drop support for older PHP versions in a new major.

PR #380, opened 2024-01-22
Upgrade to intervention/image 3

### LINKED ISSUE #377: HEIC format support
Currently Glide requires Intervention/image 2.7
According to my research it would be necessary to Intervention/image 3 to handle the HEIC format
Is there a plan for Glide to support Intervention/image 3?

### LINKED ISSUE #379: We need upgrading to intervention/image ^3 please

Art4:
This PR will drop support for PHP 7.4 and 8.0, but will allow installation of intervention/image v3. Because there will be breaking changes I propose this for Glide v3.

Refs #288.

Closes #377, #378, closes #379.


Art4:
Brief interim report

- Encoding can no longer be implemented as a manipulator, so I have moved the function to Api::run().
- interlace() and progressive jpeg are not supported anymore

Todo:

- orientate() was removed, but could be replaced by code from @konnng-dev
- rectangle() must be replaced with drawRectangle()

src/Manipulators/Background.php
```diff
@@ -26,9 +26,10 @@ public function run(Image $image)
         $color = (new Color($this->bg))->formatted();
 
         if ($color) {
-            $new = $image->getDriver()->newImage($image->width(), $image->height(), $color);
-            $new->mime = $image->mime;
-            $image = $new->insert($image, 'top-left', 0, 0);
+            $new = $image->driver()->createImage($image->width(), $image->height())->fill($color);
+            // TODO: Find a way to communicate the mime
```
ADmad, line 30: Perhaps Image::encodeByMediaType() is what we need.

ADmad: Or Image::encodeByExtension().

konnng-dev: @ADmad I was able to fix with the following approach here https://github.com/Art4/glide/pull/1/commits/f51a6278e881db70ec5c1b8a18b0e1283a8f2bba

Art4:
So, this is the current situation:

- [ ] **Encoding:** @konnng-dev has found a solution to keep Encoding implemented as a manipulator, see https://github.com/Art4/glide/pull/1
- [ ] **interlace() and progressive jpeg:** @konnng-dev has ported this functions into the Encode manipulator, see https://github.com/Art4/glide/pull/1
- [ ] **orientate()**: @konnng-dev has implemented this functions into the Orientation manipulator, see https://github.com/Art4/glide/pull/1
- [x] **rectangle()**: has been replaced with drawRectangle(), but needs more checks and bug fixes

I will finish these points and fix the tests and psalm errors to finalize this first PR.

Todo for separate PRs:

- [ ] Check if I ported the size() feature correctly to scale()
- [ ] more checks and bug fixes for border=shrink and expand

ADmad:
> Since this useful helper was removed from the official intervention v3, I wonder what we could do for this refactor to make sure it will behave as the previous version. Specially w/ fit parameter https://glide.thephpleague.com/1.0/api/size/#fit-fit

We could also consider changing Glide's API/options as per changes to Image, if trying to retain existing options is too problematic.

ADmad:
For e.g. I quite like the new "cover", "contain" naming in Image as they corelate to similar behavior in CSS.

Art4:
I'm nearly done with fixing all tests, but are struggling with Mockery on this last test.
See https://github.com/Art4/glide/actions/runs/7656671526/job/20865541487

Code: https://github.com/thephpleague/glide/blob/66ad315c190fd583e2b97206d9e87f6d056b9fac/src/Manipulators/Border.php#L141-L154

Tests: https://github.com/thephpleague/glide/blob/66ad315c190fd583e2b97206d9e87f6d056b9fac/tests/Manipulators/BorderTest.php#L92-L106

Error: `1) League\Glide\Manipulators\BorderTest::testRunOverlay
Mockery\Exception\InvalidCountException: Method width(<Any Arguments>) from Mockery_1_IteratorAggregate_Intervention_Image_Interfaces_ImageInterface should be called
 exactly 1 times but called 0 times.`

Could someone help me why the provided closure is not called?

ADmad:
@Art4 Since the `drawRectangle()` which uses the closure is mocked the closure would never actually be used/run.

src/Manipulators/Size.php
```diff
@@ -271,102 +270,93 @@ public function runResize(Image $image, $fit, $width, $height)
...
-     * @return Image The manipulated image.
+     * @return ImageInterface The manipulated image.
      */
-    public function runContainResize(Image $image, $width, $height)
+    public function runContainResize(ImageInterface $image, $width, $height): ImageInterface
     {
-        return $image->resize($width, $height, function ($constraint) {
-            $constraint->aspectRatio();
-        });
+        return $image->resize($width, $height);
```
ADmad, line 281: Should this be `return $image->scale($width, $height);`?  In v3 `Image::resize()` stretches the image and the alternative to preserve the aspect ratio is `Image::scale()` as the callback param of `resize()` is gone which allowed specifying the aspect ratio constraint.

Art4: You are right. Using `?h=600&w=500&size=contain` results in a square image. Using `scale()` know delivers the same result as before.

grafik

src/Manipulators/Size.php
```diff
@@ -271,102 +270,93 @@ public function runResize(Image $image, $fit, $width, $height)
...
+     * @return ImageInterface The manipulated image.
      */
-    public function runMaxResize(Image $image, $width, $height)
+    public function runMaxResize(ImageInterface $image, $width, $height): ImageInterface
     {
-        return $image->resize($width, $height, function ($constraint) {
-            $constraint->aspectRatio();
-            $constraint->upsize();
-        });
+        return $image->resize($width, $height);
```
ADmad, line 295: This also needs to be changed to `scaleDown()` to preserve aspect ratio and prevent upscaling.

src/Manipulators/Size.php
```diff
@@ -271,102 +270,93 @@ public function runResize(Image $image, $fit, $width, $height)
...
-     * @return Image The manipulated image.
+     * @return ImageInterface The manipulated image.
      */
-    public function runFillMaxResize(Image $image, $width, $height)
+    public function runFillMaxResize(ImageInterface $image, $width, $height): ImageInterface
     {
-        $image = $image->resize($width, $height, function ($constraint) {
-            $constraint->aspectRatio();
-        });
+        $image = $image->resize($width, $height);
```
ADmad, line 325:
```suggestion
        $image = $image->scale($width, $height);
```

src/Manipulators/Size.php
```diff
@@ -271,102 +270,93 @@ public function runResize(Image $image, $fit, $width, $height)
...
+    public function runCropResize(ImageInterface $image, $width, $height): ImageInterface
     {
         list($resize_width, $resize_height) = $this->resolveCropResizeDimensions($image, $width, $height);
 
         $zoom = $this->getCrop()[2];
 
-        $image->resize($resize_width * $zoom, $resize_height * $zoom, function ($constraint) {
-            $constraint->aspectRatio();
-        });
+        $image->resize($resize_width * $zoom, $resize_height * $zoom);
```
ADmad, line 359:
```suggestion
        $image->scale($resize_width * $zoom, $resize_height * $zoom);
```

We could probably also do away with the whole crop dimensions calculation by using the new [Image::cover()](https://image.intervention.io/v3/modifying/resizing#cropping--resizing-combined)

src/Manipulators/Size.php
Art4: Fixed. 

grafik

src/Manipulators/Size.php
Art4: Hmm, I can not find any differences using query `?h=400&w=400&size=fill-max`.

grafik

src/Manipulators/Size.php
Art4: I cannot find differences between v2 and v3 using query `?h=400&w=400&size=crop`.

grafik

src/Manipulators/Size.php
ADmad: Are these images 400x400?

src/Manipulators/Size.php
ADmad: I don't see any cropping in either :)

ADmad: You are using incorrect query string param, you need to use `fit` instead of `size`.

konnng-dev:
@Art4 I just have one objection with the current strategy to encode the image.

It seems the Encode Manipulator was commented out from `ServerFactory::getManipulators()` and moved to `Api::run`. I can see the reason why it was made.

But I still think that the Encode Manipulator is still vital for this and my suggestion is to keep that, unless @ADmad  has a different opinion.

I managed to fix the Encoded Manipulator on my branch and I was able to test it on production properly after the changes I did here https://github.com/Art4/glide/pull/2. Would you mind to check it again?

ADmad:
@konnng-dev Yes we should keep the encode manipulator since you have managed to fix it.

Art4:
> It seems the Encode Manipulator was commented out from `ServerFactory::getManipulators()` and moved to `Api::run`. I can see the reason why it was made.

Good catch. I've uncommented this, so Encode Manipulator is used again.

> But I still think that the Encode Manipulator is still vital for this and my suggestion is to keep that, unless @ADmad has a different opinion.
> 
> I managed to fix the Encoded Manipulator on my branch and I was able to test it on production properly after the changes I did here [Art4#2](https://github.com/Art4/glide/pull/2). Would you mind to check it again?

I would recommend to create a separate PR in this main repo about this issue, instead in my private repo, to keep the version history and discussions in this repo.

src/Manipulators/Size.php
Art4: :facepalm: :sweat_smile:

ADmad:
I still have to do a more thorough review of the PR but I am going to merge it so that @konnng-dev can submit his changes on top.

Thanks guys for all your work. Please start trying out the 3.x branch as due to the nature of this lib passing tests are no guarantee that the images generated are what we expect 😄.

PR #387, opened 2024-01-30
Fix interlacing

ADmad:
@konnng-dev I took your changes from https://github.com/Art4/glide/pull/2 to fix interlacing.

Though when testing the generated image, interlacing doesn't seem to work for JPEGs when using GD.
With Imagick `identify -verbose kayak.jpeg | grep Interlace` gives `Interlace: JPEG` but with GD it gives `Interlace: None`.
It seems to work fine for PNGs for either driver.

@Art4 BTW Glide v2 doesn't interlace PNGs but you have added interlacing for all PNGs in v3. Is that intentional?

### DIFF
--- a/src/Manipulators/Encode.php
+++ b/src/Manipulators/Encode.php
@@ -5,6 +5,7 @@ namespace League\Glide\Manipulators;
 use Intervention\Image\Drivers\Gd\Driver as GdDriver;
 use Intervention\Image\Drivers\Imagick\Driver as ImagickDriver;
 use Intervention\Image\ImageManager;
+use Intervention\Image\Interfaces\DriverInterface;
 use Intervention\Image\Interfaces\ImageInterface;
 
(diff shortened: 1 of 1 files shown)


ADmad:
> Interlacing for all png was added by @konnng-dev in https://github.com/Art4/glide/pull/1

Hmm.. I am leaning towards reverting it and maintaining status quo.

konnng-dev:
Oh, it was indeed my fault 

I can submit a new PR just interlacing JPEGS. But I would like to have option to interlace pngs on demand as well.

ADmad:
> But I would like to have option to interlace pngs on demand as well.

You can open a separate PR for that.

ADmad:
Gonna merge this. Will look into JPG interlacing not working with Gd later or just state that as a limitation if me or someone else can't sort it out.

Issue #153, opened 2016-08-25
Security issue with md5


Hacker12345577342:
Nowaday there are a lot of tools to decrypte md5 which Glide use for sign on images.when people decrypte one of them they will see our secret key  because Glide generate key like this 
return md5($this->signKey.':250182131287'.ltrim($path, '/').'?'.http_build_query($params));(I added before some numbers but it doesnt matter there).If someone decrypt key they will my secret key,then they will try change sizes (maybe with some script to automatic generated sizes) and then they will md5 this then they will send this key to Glide server,then Glide server will try to handle a lot of cache and will crash

Alex12341232:
Wow it is true ,I can also do this with online decryptes

francescolaffi:
this security concern is probably far fetched in 99.99% of use cases, however it looks like `https://github.com/thephpleague/glide/blob/master/src/Signatures/Signature.php` is trivial to override the `generateSignature` method to use a more secure hashing algo if you deem it necessary

just imho, im not a maintainer of this lib

reinink:
First of, yes, this can be an issue, which is why we [highly recommend](http://glide.thephpleague.com/1.0/config/security/) using a large signing key.

![image](https://cloud.githubusercontent.com/assets/882133/18004509/5e82230a-6b62-11e6-85d1-9d0f83173277.png)

The reason we use md5 is because it's fast. These request happen in realtime, so have a 1+ second encryption delay isn't very practical. That said, there isn't any reason why you couldn't extend the signature class as @francescolaffi is suggesting and add in your preferred signature generation technique.

At this point I'd probably like to get some input from some security experts who know more about this stuff than I do. I'm going to reach out to a couple folks.

I'd like to also see what other libraries that offer similar functionality are doing. Here are a couple examples to start:

- Twilio uses `hash_hmac` ([source](https://github.com/twilio/twilio-php/blob/master/Twilio/Security/RequestValidator.php#L28))
- Imgix uses `crc32` ([source](https://github.com/imgix/imgix-php/blob/master/src/Imgix/UrlBuilder.php#L69))

paragonie-scott:
https://github.com/thephpleague/glide/blob/5d71f2fb030907fdd9552b4952d1ecdc1804702e/src/Signatures/Signature.php#L61

You want `hash_hmac()`. The current construction is vulnerable to [length-extension attacks](https://blog.skullsecurity.org/2012/everything-you-need-to-know-about-hash-length-extension-attacks).

[...]

PR #396, opened 2024-05-17
3.x-use-xxh3

nlemoine:
Since Glide now requires php ^8.1, replace `md5` with `xxh3` for shorter and faster hash, see: https://php.watch/versions/8.1/xxHash

Note this is a breaking change that will trigger all images to be recreated.

Maybe that could be introduced behind an optin flag?

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -349,7 +349,7 @@ class Server
         unset($params['s'], $params['p']);
         ksort($params);
 
-        $cachedPath = md5($sourcePath.'?'.http_build_query($params));
+        $cachedPath = hash('xxh3', $sourcePath.'?'.http_build_query($params));
 
         if ($this->groupCacheInFolders) {
             $cachedPath = $sourcePath.'/'.$cachedPath;
--- a/tests/ServerTest.php
+++ b/tests/ServerTest.php
@@ -234,7 +234,7 @@ class ServerTest extends TestCase
     public function testGetCachePath()
     {
         $this->assertEquals(
-            'image.jpg/e863e008b6f09807c3b0aa3805bc9c63',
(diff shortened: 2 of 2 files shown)


ADmad:
There have been concerns in the past regarding the use of md5 as stated in this issue #153, so I am fine changing it.

While we are at it the code should also be updated as per suggestions in this comment https://github.com/thephpleague/glide/issues/153#issuecomment-242756953. Though I am not sure if `hash_hmac()` supports `xxh3`.

I also don't want to inconvenience existing users who are upgrading, so it would be nice to provide an option to continue using the existing hashing mechanism.

nlemoine:
@ADmad Oh wait, just to clarify, my PR was only targeting the hashing algorithm used to handle unique file names. It was not about signing requests (I'm only using the API part of Glide, e.g. not the "server" part). 

As it's stated in the link posted above:
> xxHash is not a cryptographic hashing algorithm. For password hashing, use [password_hash](https://www.php.net/manual/function.password-hash.php) and its friends. Furthermore, none of the xxHash variants are allowed in [hash_hmac](https://www.php.net/manual/function.hash-hmac.php) function.

xxh is not meant to provide security, just fast hashing.

ADmad:
Oops sorry, I totally misread the PR. Shorter and faster hashes for filenames is great.

Given that this lib is for generating images "on the fly" I don't think having all the cached images regenerated is an issue. We can document this change in the release notes so that people can clean up existing cached images and/or handle cache warm up themselves.

nlemoine:
> I don't think having all the cached images regenerated is an issue

Depends how much images we're talking about :) I personally don't bother. I'll let you merge this if this ok the way it is then.

PR #410, opened 2024-12-16
Encode using media type instead of extension

deluxetom:
If an image had `image/x-webp` or another format supported by `Intervention` but not defined in `supportedFormats`, it would fail.

This uses `encodeByMediaType` instead which uses the proper values and available formats defined by `Intervention`

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -26,8 +26,8 @@ class Api implements ApiInterface
     /**
      * Create API instance.
      *
-     * @param ImageManager                    $imageManager Intervention image manager.
-     * @param list<ManipulatorInterface|null> $manipulators Collection of manipulators.
+     * @param ImageManager               $imageManager Intervention image manager.
+     * @param list<ManipulatorInterface> $manipulators Collection of manipulators.
      */
     public function __construct(ImageManager $imageManager, array $manipulators)
     {
@@ -59,6 +59,8 @@ class Api implements ApiInterface
      * Set the manipulators.
      *
      * @param array $manipulators Collection of manipulators.
+     *
+     * @throws \InvalidArgumentException
      */
     public function setManipulators(array $manipulators): void
     {
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -27,7 +27,7 @@ class Api implements ApiInterface
      * Create API instance.
      *
      * @param ImageManager $imageManager Intervention image manager.
-     * @param array        $manipulators Collection of manipulators.
+     * @param list<ManipulatorInterface|null> $manipulators Collection of manipulators.
      */
     public function __construct(ImageManager $imageManager, array $manipulators)
     {
@@ -58,17 +58,11 @@ class Api implements ApiInterface
     /**
      * Set the manipulators.
      *
-     * @param ManipulatorInterface[] $manipulators Collection of manipulators.
+     * @param list<ManipulatorInterface|null> $manipulators Collection of manipulators.
      */
     public function setManipulators(array $manipulators): void
     {
-        foreach ($manipulators as $manipulator) {
-            if (!($manipulator instanceof ManipulatorInterface)) {
-                throw new \InvalidArgumentException('Not a valid manipulator.');
-            }
-        }
-
-        $this->manipulators = $manipulators;
+        $this->manipulators = array_filter($manipulators, fn ($manipulator) => $manipulator instanceof ManipulatorInterface);
     }
 
     /**
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -26,7 +26,7 @@ class Api implements ApiInterface
     /**
      * Create API instance.
      *
-     * @param ImageManager $imageManager Intervention image manager.
+     * @param ImageManager                    $imageManager Intervention image manager.
      * @param list<ManipulatorInterface|null> $manipulators Collection of manipulators.
      */
     public function __construct(ImageManager $imageManager, array $manipulators)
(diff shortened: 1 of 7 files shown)


src/Api/Encoder.php
```diff
@@ -101,35 +108,68 @@ public function run(ImageInterface $image): EncodedImageInterface
      */
     public function getFormat(ImageInterface $image): string
     {
-        $fm = (string) $this->getParam('fm');
```
ADmad, line 104: This needs to be preserved. If `fm` is specified then that's the format that should be returned irrespective of the actual media type.

src/Api/Encoder.php
```diff
@@ -101,35 +108,68 @@ public function run(ImageInterface $image): EncodedImageInterface
...
-
-        /** @psalm-suppress RiskyTruthyFalsyComparison */
-        return array_search($image->origin()->mediaType(), static::supportedFormats(), true) ?: 'jpg';
     }
 
     /**
      * Get a list of supported image formats and MIME types.
      *
-     * @return array<string,string>
+     * @return array<string,MediaType>
```
ADmad, line 124: Changing the type can cause problems for users upgrading from 2.x. I prefer not breaking backwards compatibility unless absolutely necessary.

tests/Api/EncoderTest.php
```diff
@@ -102,53 +115,43 @@ public function testRun(): void
...
-
-            if (function_exists('imagecreatefromwebp')) {
-                $this->assertMediaType($mock, 'image/webp')->once();
-            }
-
-            if (function_exists('imagecreatefromavif')) {
-                $this->assertMediaType($mock, 'image/avif')->once();
-            }
-        });
+        $this->assertSame('jpg', $this->encoder->setParams(['fm' => 'jpg'])->getFormat($this->getImageByMimeType('image/jpeg')));
```
ADmad, line 124: We also need tests where `fm` does not match the image's mime type. `fm`'s value should take precedence in that case.

src/Api/Encoder.php
deluxetom: it's preserved, its just moved there:
https://github.com/thephpleague/glide/pull/410/files#diff-63a111a0d299d1ea5fa389c1909b9577ed17e30e75c26627802b3cb3c8cdd606R89

tests/Api/EncoderTest.php
deluxetom: I added more tests to check the `fm` parameter always takes precedence

src/Api/Encoder.php
deluxetom: fixed

src/Api/Encoder.php
ADmad: Well directly calling this method won't take into account `fm` would it?  This is a public API.

ADmad: Sorry, I misread the code, `fm`'s value is accounted for.

src/Api/Api.php
```diff
@@ -58,17 +58,11 @@ public function getImageManager(): ImageManager
     /**
      * Set the manipulators.
      *
-     * @param ManipulatorInterface[] $manipulators Collection of manipulators.
+     * @param array $manipulators Collection of manipulators.
      */
```
ADmad, line 62:
```suggestion
     * @throws \InvalidArgumentException
     */
```

src/Api/Api.php
```diff
@@ -26,8 +26,8 @@ class Api implements ApiInterface
     /**
      * Create API instance.
      *
-     * @param ImageManager $imageManager Intervention image manager.
-     * @param array        $manipulators Collection of manipulators.
+     * @param ImageManager                    $imageManager Intervention image manager.
+     * @param list<ManipulatorInterface|null> $manipulators Collection of manipulators.
```
ADmad, line 30: I don't think `|null` is necessary

deluxetom: agreed, its been fixed

src/Api/Api.php
deluxetom: @ADmad good catch,  I added the missing phpdoc

ADmad:
Something went wrong, testsuite is failing.

deluxetom:
> Something went wrong, testsuite is failing.

GitHub is having some issues, I can't restart the workflow, tests are all green locally:
```
> ./vendor/bin/phpunit
PHPUnit 11.5.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.1
Configuration: phpunit.xml.dist

...............................................................  63 / 216 ( 29%)
............................................................... 126 / 216 ( 58%)
............................................................... 189 / 216 ( 87%)
...........................                                     216 / 216 (100%)

Time: 00:00.327, Memory: 16.00 MB

OK (216 tests, 455 assertions)
```

I'll try to fix as soon as I can

deluxetom:
@ADmad it seems `heic` isn't supported anymore, I'm not sure I understand why, it started when I merged the `3.x` branch from upstream.

ADmad:
Aha, I remember now, `heic` decoding is broken on ubuntu-24.04. I forgot about that and updated the github actions config to use 24.04 instead of 22.04. Please revert the image for `tests` job to `ubuntu-22.04` and the error will be gone.

deluxetom:
@ADmad all green now, let me know if there's anything you want me to update

ADmad:
Looks good to go. I'll just have a look again with fresh eyes tomorrow before merging.

ADmad:
@deluxetom I re-thought about the problem your PR aims to address and come up with an alternative which require much lesser code changes.

ADmad:
Closing as the alternative PR is merged. Thank you @deluxetom.

PR #413, opened 2024-12-17
Define API params in order to be able to filter correctly

### LINKED ISSUE #412: All parameters are used for the cache path hash
- `image.jpg?w=300&h=400&fit=crop`
- `image.jpg?w=300&h=400&fit=crop&invalid=1` will have a different cache path when it shouldn't

I'm working on a fix that checks the parameters passed to the Glide server and API, I'll post a PR soon

deluxetom:
* resolves #412 

- Added new tests to make sure only valid parameters are used
- New interface method for each manipulator to know the API parameters used
- no BC break

### DIFF
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -11,6 +11,13 @@ use League\Glide\Manipulators\ManipulatorInterface;
 
 class Api implements ApiInterface
 {
+    public const GLOBAL_API_PARAMS = [
+        'p', // presets
+        'q', // quality
+        'fm', // format
+        's', // signature
+    ];
+
     /**
      * Intervention image manager.
      */
@@ -24,6 +31,13 @@ class Api implements ApiInterface
     protected array $manipulators;
 
     /**
+     * API parameters.
+     *
+     * @var list<string>
+     */
+    protected array $apiParams;
+
+    /**
      * Create API instance.
      *
      * @param ImageManager $imageManager Intervention image manager.
@@ -33,6 +47,7 @@ class Api implements ApiInterface
     {
         $this->setImageManager($imageManager);
         $this->setManipulators($manipulators);
+        $this->setApiParams();
     }
 
     /**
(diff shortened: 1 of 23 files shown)


tests/ServerTest.php
```diff
@@ -410,6 +409,7 @@ public function testGetAllParams()
         $all_params = $this->server->getAllParams([
             'w' => '100',
             'p' => 'small',
+            'invalid' => '1',
```
deluxetom, line 412: @ADmad I realize I added some fixes that are out of scope of the issue, but this is the test fixing the cache path issue

I believe it shouldn't accept invalid parameters

src/Api/Api.php
```diff
@@ -11,6 +11,13 @@
 
 class Api implements ApiInterface
 {
+    public const GLOBAL_API_PARAMS = [
+        'p', // presets
```
ADmad, line 15:
```suggestion
        'p', // preset
```

src/Api/Api.php
```diff
@@ -116,4 +131,32 @@ public function encode(ImageInterface $image, array $params): string
...
+     * Sets the API parameters for all manipulators.
+     *
+     * @return list<string>
+     */
+    public function setApiParams(): array
+    {
+        $params = [];
+
+        array_walk($this->manipulators, function (ManipulatorInterface $manipulator) use (&$params) {
+            foreach ($manipulator->getApiParams() as $param) {
```
ADmad, line 145: Why loop through each instead of using array merge?

src/Manipulators/BaseManipulator.php
```diff
@@ -37,6 +37,13 @@ public function getParam(string $name): mixed
             : null;
     }
 
+    /**
+     * Get the names of the manipulator API parameters.
+     *
+     * @rerun list<string>
```
ADmad, line 43:
```suggestion
     * @return list<string>
```

ADmad (commented):
> no BC break

Adding a new interface method is very much a BC break :slightly_smiling_face: 

But I think it's worth it in this case, given the improvement it provides.

src/Manipulators/BaseManipulator.php
```diff
@@ -37,6 +37,13 @@ public function getParam(string $name): mixed
             : null;
     }
 
+    /**
+     * Get the names of the manipulator API parameters.
+     *
+     * @return list<string>
+     */
+    abstract public function getApiParams(): array;
```
Art4, line 45: We don't need this abstract method because it is defined already in ManipulatorInterface

src/Api/Api.php
```diff
@@ -116,4 +131,32 @@ public function encode(ImageInterface $image, array $params): string
...
+     * @return list<string>
+     */
+    public function setApiParams(): array
+    {
+        $this->apiParams = self::GLOBAL_API_PARAMS;
+
+        foreach ($this->manipulators as $manipulator) {
+            foreach ($manipulator->getApiParams() as $param) {
+                $this->apiParams[] = $param;
+            }
```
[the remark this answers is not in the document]
deluxetom: I was over-complicating this, thanks for pointing it out!

src/Manipulators/BaseManipulator.php
deluxetom: I agree, I was following what was done for the `run` method, I removed both from `BaseManipulator`?

PR #417, opened 2024-12-19
Local dev improvements

deluxetom:
This made my life much easier in the last couple of days, so I wanted to share.

Some dependencies are only for `php-vips`, I'm currently working on getting the `intervention` driver for `vips` ready.
I would love for Glide to also support the `vips` driver as it is very fast and uses fewer resources


composer.json
```diff
@@ -34,11 +34,17 @@
     "require-dev": {
         "mockery/mockery": "^1.6",
         "phpunit/phpunit": "^10.5 || ^11.0",
-        "friendsofphp/php-cs-fixer": "^3.48"
+        "friendsofphp/php-cs-fixer": "^3.48",
+        "phpstan/phpstan": "^1.10 || ^2.0"
```
ADmad, line 38:
```suggestion
        "phpstan/phpstan": "^2.0"
```

v1 vs v2 can generate slightly different error report

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM php:8.1-cli
+
+RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
+
+# install dependencies
+RUN apt update && apt install -y --no-install-recommends \
+        libvips42 \
```
ADmad, line 7: Let's hold off the dependencies for vips until we are able to actually support it.

ADmad (commented):
`.gitattributes` needs to the updated to "export-ignore" the new composer config files.

deluxetom:
@ADmad PR updated

PR #418, opened 2024-12-20
Fixed param typo

deluxetom:
Had a typo for an API param 

I wonder if we should throw an exception when a manipulator gets a param that's not allowed?


ADmad:
> I wonder if we should throw an exception when a manipulator gets a param that's not allowed?

Well `Api::run()` calls that the `setParams()` of each manipulator before running it. If we were to add an exception we would have to first get the valid params list of each manipulator and filter out the "invalid" ones. I don't think there's anything to be gained by doing that, it would just add CPU cycles.

PR #419, opened 2025-01-02
Use `cover` method from Intervention

### LINKED ISSUE #420: `fit=crop` should use the resize `cover` method from Intervention
When possible, `Glide` should use the built-in methods from `Intervention`:
https://image.intervention.io/v3/modifying/resizing#cropping--resizing-combined

I noticed differences when `fit=crop` was used, the way the crop is calculated in `glide` can create 1px differences.

deluxetom:
* resolves #420

Intervention v3 has a `cover` method that works as intended for the `fit=crop` parameter.
https://image.intervention.io/v3/modifying/resizing#cropping--resizing-combined

### DIFF
--- a/docs/3.0/api/size.md
+++ b/docs/3.0/api/size.md
@@ -37,6 +37,7 @@ Sets how the image is fitted to its target dimensions.
 - `fill-max`: Resizes the image to fit within the width and height boundaries without cropping but upscaling the image if it's smaller. The finished image will have remaining space on either width or height (except if the aspect ratio of the new image is the same as the old image). The remaining space will be filled with the background color. The resulting image will match the constraining dimensions.
 - `stretch`: Stretches the image to fit the constraining dimensions exactly. The resulting image will fill the dimensions, and will not maintain the aspect ratio of the input image.
 - `cover`: Resizes the image to fill the width and height boundaries and crops any excess image data. The resulting image will match the width and height constraints without distorting the image. See the [crop](api/crop/) page for more information.
+- `crop`: Resizes the image to fill the width and height boundaries and crops any excess image data. (alias for `cover`).
 - `crop-x%-y%`: Resizes the image to fill the width and height boundaries and crops based on a focal point defined by `x%` (left offset) and `y%` (top offset). The resulting image will match the width and height constraints without distorting the image. See the [crop](api/crop/) page for more information.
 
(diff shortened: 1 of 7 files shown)


deluxetom:
@ADmad I've been using this update with a project already, the documentation for `glide` could also be updated to add `fit=cover`

ADmad:
> the documentation for glide could also be updated to add fit=cover

Sounds good to me, would you like to take care of that?

deluxetom:
> > the documentation for glide could also be updated to add fit=cover
> 
> Sounds good to me, would you like to take care of that?

I think I can get started on the `3.0` docs, it looks like its not done yet

docs/3.0/api/size.md
```diff
@@ -29,14 +29,15 @@ Sets the height of the image, in pixels.
...
 - `fill`: Resizes the image to fit within the width and height boundaries without cropping or distorting the image, and the remaining space is filled with the background color. The resulting image will match the constraining dimensions.
 - `fill-max`: Resizes the image to fit within the width and height boundaries without cropping but upscaling the image if it's smaller. The finished image will have remaining space on either width or height (except if the aspect ratio of the new image is the same as the old image). The remaining space will be filled with the background color. The resulting image will match the constraining dimensions.
 - `stretch`: Stretches the image to fit the constraining dimensions exactly. The resulting image will fill the dimensions, and will not maintain the aspect ratio of the input image.
-- `crop`: Resizes the image to fill the width and height boundaries and crops any excess image data. The resulting image will match the width and height constraints without distorting the image. See the [crop](api/crop/) page for more information.
```
ADmad, line 39: I think we should keep the `crop` option stating it's an alias for `cover`.

deluxetom:
> Just to confirm using `fit=crop` or `fit=crop-<string>-<string>` would still work right? Internally `runCoverResize()` would be used.
> 
> `runCropReize()` would be used only `fit=crop-<numeric>-<numeric>` is used, right?

that's correct, `crop` and `crop-$position` will use `runCoverResize()` now, they're aliases of `cover` and `cover-position`.
`crop-<numeric>-<numeric>` remains the same

ADmad:
@deluxetom Can you please address the couple of comments I had left for the docs page updates.

deluxetom:
@ADmad updated!

PR #427, opened 2025-03-21
3.x

ADmad:


duncanmcclean:
Thanks for your work on Glide 3!

Is there an upgrade guide anywhere detailing any changes in behaviour, or should everything "just work' after updating the version constraint?

I couldn't find one on the docs, or on the changelog.

ADmad:
Here's the changelog https://glide.thephpleague.com/3.0/changelog/

In general everything will "just work" after updating the constraint. Only if you have a custom manipulator it will need updating.

duncanmcclean:
Great, thank you for clarifying!

PR #428, opened 2025-03-21
Simplify Background::class manipulator

olivervogel:
For the calls of the [Background `bg`](https://glide.thephpleague.com/3.0/api/background/) manipulator Intervention Image has already an own method [`blendTransparency()`](https://image.intervention.io/v3/basics/colors#merge-transparent-areas-with-color) to set the background color of the image. 

This patch simplifies the `Background::class` manipulator in that way and replaces the manual individual steps with the mentioned method.

### DIFF
--- a/src/Manipulators/Background.php
+++ b/src/Manipulators/Background.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace League\Glide\Manipulators;
 
 use Intervention\Image\Interfaces\ImageInterface;
-use Intervention\Image\Origin;
 use League\Glide\Manipulators\Helpers\Color;
 
 class Background extends BaseManipulator
@@ -32,11 +31,6 @@ class Background extends BaseManipulator
 
         $color = (new Color($bg))->formatted();
 
-        return $image->driver()->createImage($image->width(), $image->height())
-            ->fill($color)
-            ->place($image, 'top-left', 0, 0)
-            ->setOrigin(
-                new Origin($image->origin()->mediaType())
-            );
+        return $image->blendTransparency($color);
     }
 }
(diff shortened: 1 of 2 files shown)


deluxetom (approved):

ADmad:
While looking into this I found that setting `bg` doesn't work when for e.g. using `fit=fill` and the using `w` and `h` larger than the original image dimensions. The extra area if always white since the default background color for `Image::pad()` is `ffffff`. Before merging this change, changing the background color to `'ffffff00'` for `Image::pad()` fixed this issue but after merging this the background color is "skewed".

Before merging this PR using `bg=blue` I get
k1
After this merge
k2

ADmad:
Problem seems to occur only when blue color is involved. If I use `red` or `green` the results appear same before and after this patch.

olivervogel:
I think the definition of the default background color in `Image::pad()` and other methods could have been implemented more logically in Intervention Image. Unfortunately a breaking change.

See https://github.com/Intervention/image/issues/1391#issuecomment-2745146586

ADmad:
Yes, `ffffff` does not seem like a good default, `null` or `transparent` would be better.

The problem I described above only occurs when using `imagick`, not `gd`. So to summarize, when using `imagick` and `blue` as the blending color, the resulting background color is off (PHP 8.4 and imagick 3.7.0 ).

olivervogel:
> The problem I described above only occurs when using `imagick`, not `gd`. So to summarize, when using `imagick` and `blue` as the blending color, the resulting background color is off (PHP 8.4 and imagick 3.7.0 ).

I'm not sure but maybe it's better to revert my changes.

ADmad:
I'll try to look into a bit more over the weekend and then decide. Will also try to submit a test case / minimal example for the issue.

ADmad:
I am going to revert this. The blending color is only used when outputting to image format which does not support transparency. But when `bg` is specified that color should replaced the transparent area.

PR #443, opened 2026-02-12
Make the kayak great again

### LINKED ISSUE #328: Heroku preview app is dead

guillaume-sainthillier:
This PR generate and self host the images used in documentation in order to fix #328 for free.

a php script (included) was used to generate the image and store it to the /docs folder.

I've tested locally in every documentation branch and it's OK.

For me the main caveat is that the API is not clearly exposed as it should be, making possibly confusion for users looking for the _real_ query parameters, this is why I kept them in the local image URIs


ADmad:
Thanks, could you update `.gitattributes` file too to export-ignore the `scripts` folder.

guillaume-sainthillier:
> Thanks, could you update `.gitattributes` file too to export-ignore the `scripts` folder.

Done :)

PR #446, opened 2026-03-28
Migrate to Intervention Image v4

olivervogel:
This PR makes the necessary changes to migrate to the just released [Intervention Image Version 4](https://github.com/Intervention/image/releases/tag/4.0.0).

Please note that the new version requires at least PHP 8.3.

The docker test runner currently fails with an error, but this was already happening before I made any changes. Unfortunately, I can't say for sure what's causing it.

However, I think this is a good start and takes care of most of the work for the update. I recommend a thorough review.

If you have any questions, feel free to reach out to me.

PS: For your reference, here is the [upgrade guide](https://image.intervention.io/v4/getting-started/upgrade)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -12,10 +12,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['8.1', '8.2', '8.3', '8.4', '8.5']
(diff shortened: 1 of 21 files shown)


Art4:
Great work. 💪

deluxetom (approved):

PR #447, opened 2026-03-28
Throw an exception if an invalid format is provided.

### LINKED ISSUE #416: RFC - Throw an exception if an invalid value format is provided instead of silently converting it to `jpg`.
Currently any invalid value for `fm` gets silently coerced to `jpg` by the encoder. This behavior, while ensure an image is always generated, has couple of issues.

- A dev error like a typo in the format name would be difficult to track down since the lib gives no error about it.
- Since all parameter values are used to generate the hash key using multiple invalid value for `fm` would generate multiple cached images which would be actually the same.

Based on the community feedback I would like to throw an exception instead of invalid format value.

Tagging in recent contributors to the 3.x branch @deluxetom @Art4 @nlemoine @konnng-dev

ADmad:
Closes #416

/cc @Art4 @deluxetom

### DIFF
--- a/src/Api/Encoder.php
+++ b/src/Api/Encoder.php
@@ -104,7 +104,11 @@ class Encoder
     {
         $fm = (string) $this->getParam('fm');
         if ($fm) {
-            return array_key_exists($fm, static::supportedFormats()) ? $fm : 'jpg';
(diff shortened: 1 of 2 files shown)


Art4 (approved):
👏

PR #450, opened 2026-04-02
Fix: normalize watermark alpha to 0-1 range for Intervention Image v4

### LINKED ISSUE #449: Bug: `Watermark::getAlpha()` returns 0-100 range but Intervention Image v4 expects 0-1
### Description

`Watermark::getAlpha()` returns an `int` in the 0-100 range (defaulting to `100`), but Intervention Image v4's `InsertModifier` expects a `float` in the 0-1 range. This causes an `InvalidArgumentException` on every request that applies a watermark.

### Versions

- `league/glide`: `4.0.0-beta1`
- `intervention/image`: `4.0.0`

### Error

```
Intervention\Image\Exceptions\InvalidArgumentException: Transparency must be in range 0 to 1
  at vendor/intervention/image/src/Modifiers/InsertModifier.php:26
```

### Root Cause

In [`src/Manipulators/Watermark.php:265-278`](https://github.com/thephpleague/glide/blob/4.x/src/Manipulators/Watermark.php#L265-L278):

```php
public function getAlpha(): int
{
    $markalpha = $this->getParam('markalpha');

    if (!is_numeric($markalpha)) {
        return 100;
    }

    if ($markalpha < 0 || $markalpha > 100) {
        return 100;
    }

    return (int) $markalpha;
}
```

This value is passed directly to `Image::insert()` on [line 117](https://github.com/thephpleague/glide/blob/4.x/src/Manipulators/Watermark.php#L117):

```php
return $image->insert($watermark, intval($markx), intval($marky), $markpos, $markalpha);
```

However, Intervention Image v4's `InsertModifier::__construct()` validates that transparency is between 0 and 1:

```php
public function __construct(
    public mixed $image,
    public int $x = 0,
    public int $y = 0,
    public string|Alignment $alignment = Alignment::TOP_LEFT,
    public float $transparency = 1
) {
    if ($this->transparency < 0 || $this->transparency > 1) {
        throw new InvalidArgumentException('Transparency must be in range 0 to 1');
    }
}
```

### Suggested Fix

Normalize the 0-100 integer to a 0-1 float in `getAlpha()`:

```php
public function getAlpha(): float
{
    $markalpha = $this->getParam('markalpha');

    if (!is_numeric($markalpha)) {
        return 1.0;
    }

    if ($markalpha < 0 || $markalpha > 100) {
        return 1.0;
    }

    return (float) $markalpha / 100;
}
```

### Steps to Reproduce

1. Configure a Glide server with a watermark filesystem
2. Set a default `mark` parameter (no `markalpha` override needed — the default `100` triggers it)
3. Request any image — the watermark manipulator throws on every request

deluxetom:
## Summary

- Fixes `Watermark::getAlpha()` returning 0-100 range while Intervention Image v4's `InsertModifier` expects 0-1 float
- Normalizes the value at the call site in `run()` (`$markalpha / 100`) to maintain backwards compatibility for `getAlpha()`
- No changes to the public API — `getAlpha()` still returns `int` in 0-100 range

Closes #449

## Test plan

- [x] `testGetAlpha` assertions unchanged and passing
- [ ] Manual test: configure a Glide server with watermark filesystem and verify watermarks render without `InvalidArgumentException`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Manipulators/Watermark.php
+++ b/src/Manipulators/Watermark.php
@@ -114,7 +114,7 @@ class Watermark extends BaseManipulator
         ]);
         $watermark = $size->run($watermark);
 
-        return $image->insert($watermark, intval($markx), intval($marky), $markpos, $markalpha);
+        return $image->insert($watermark, intval($markx), intval($marky), $markpos, $markalpha / 100);
     }
 
     /**
--- a/tests/Manipulators/WatermarkTest.php
+++ b/tests/Manipulators/WatermarkTest.php
@@ -222,7 +222,7 @@ class WatermarkTest extends TestCase
         $image = \Mockery::mock(ImageInterface::class, function ($mock) use ($watermarkImage) {
             $mock->shouldReceive('insert')
                 ->withArgs(function ($wm, $x, $y, $pos, $alpha) {
-                    return $alpha >= 0 && $alpha <= 1 && $alpha === 0.65;
+                    return $alpha >= 0 && $alpha <= 1 && 0.65 === $alpha;
                 })
                 ->once()
                 ->andReturnSelf();
--- a/tests/Manipulators/WatermarkTest.php
+++ b/tests/Manipulators/WatermarkTest.php
@@ -212,6 +212,43 @@ class WatermarkTest extends TestCase
         $this->assertSame('bottom-right', $this->manipulator->setParams(['markpos' => 'invalid'])->getPosition());
     }
 
+    public function testRunNormalizesAlphaToZeroOneRange()
+    {
+        $watermarkImage = \Mockery::mock(ImageInterface::class, function ($mock) {
+            $mock->shouldReceive('width')->andReturn(0)->once();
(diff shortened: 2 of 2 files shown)


ADmad:
> testGetAlpha assertions unchanged and passing

That doesn't do much good does it, since they were passing earlier too with incorrect code :)

deluxetom:
@ADmad I wanted to keep it backwards compatible, I can update it so `testGetAlpha` returns a value between `0` and `1`, let me know

ADmad:
We should keep backwards compatibility. What I meant was that current tests are not enough since they didn't catch this bug. But asserting the values sent to `Image::insert()` would mean using a mocked instance.

deluxetom:
@ADmad regression test added

deluxetom:
@ADmad friendly bump :) let me know if you require anymore changes

PR #455, opened 2026-07-14
Fix 1px background line when using focal point cropping

### LINKED ISSUE #451: Rounding issue when using focal point cropping results in 1px white line
When using focal point cropping (e.g. test.jpg?w=600&h=250&fit=crop-33-66), depending on the aspect ratio and dimensions of the original image, in some cases the cropped result contains a one pixel white line on either edge.

## Cause

This happens, because before the image is cropped, it is scaled down to the given dimensions while keeping aspect ratio. Width and height are rounded (see  [Manipulators/Size.php](https://github.com/thephpleague/glide/blob/master/src/Manipulators/Size.php#L348)) - which as a result, can be less than the width or height that are passed to cropping, causing the crop offset to be wrong and a pixel of the white background color to not be covered by the image.

```php
$image->scale((int) round($resize_width * $zoom), (int) round($resize_height * $zoom));

[$offset_x, $offset_y] = $this->resolveCropOffset($image, $width, $height);
// ^-- $width (or $height) can differ by 1 to width and height of the scaled image, 
// which could be rounded down, resulting in $offset_x being -1

return $image->crop($width, $height, $offset_x, $offset_y, 'transparent');
```
## Example 

Here's an example of a 1920x1278 source image:

Image

And here's the result, cropped with the following parameters:

test.jpg?w=600&h=250&fit=crop-0-50

Image

## Suggested fix

Simply changing the rounding function for parameters passed to the scaling method mentioned above to `ceil()` instead of `round()` fixes the issue, because then the scaled image will have at least the size of the given crop dimensions.

```php
$image->scale((int) ceil($resize_width * $zoom), (int) ceil($resize_height * $zoom));
```

lazerg:
Focal point cropping rounds the resize dimensions down, so the scaled image can end up a pixel short of the crop and leaves a 1px background line along an edge. Rounding up instead makes the scaled image fully cover the crop.

Fixes #451

### DIFF
--- a/src/Manipulators/Size.php
+++ b/src/Manipulators/Size.php
@@ -346,7 +346,7 @@ class Size extends BaseManipulator
 
         $zoom = $this->getCrop()[2];
 
-        $image->scale((int) round($resize_width * $zoom), (int) round($resize_height * $zoom));
+        $image->scale((int) ceil($resize_width * $zoom), (int) ceil($resize_height * $zoom));
 
         [$offset_x, $offset_y] = $this->resolveCropOffset($image, $width, $height);
 
--- a/tests/Manipulators/SizeTest.php
+++ b/tests/Manipulators/SizeTest.php
@@ -268,6 +268,23 @@ class SizeTest extends TestCase
         );
     }
 
+    public function testRunCropResizeScalesUpToFullyCoverCropDimensions(): void
+    {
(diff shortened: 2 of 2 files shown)


lazerg:
hey @ADmad, mind taking a look when you have a moment?

PR #457, opened 2026-07-16
Add a strip parameter to control EXIF metadata

### LINKED ISSUE #445: Expose Intervention/Image strip option in Glide for consistent EXIF handling
### Description

Hi,

Glide uses Intervention/Image internally to process images, which itself relies on either **GD** or **Imagick**.

I noticed an inconsistency in EXIF metadata handling depending on the driver used.

### Current behavior

- GD (via Intervention/Image)
  EXIF metadata are always stripped, with no way to keep them.
- Imagick (via Intervention/Image)
  - If the strip option is not provided, all EXIF metadata are preserved
  - If strip = true, Intervention/Image removes EXIF metadata while keeping color profiles 

This means that simply switching the driver from GD to Imagick can unexpectedly reintroduce EXIF metadata.

### Problem

Glide currently does not expose or forward the `strip` parameter to Intervention/Image.
As a result, users cannot explicitly request EXIF stripping when using the Imagick driver, even though Intervention/Image already supports this behavior.

### Expected / suggested behavior

Glide should allow passing a strip parameter to Intervention/Image so users can control metadata handling consistently across drivers.

This would:

- Enable EXIF stripping with Imagick
- Preserve color profiles when stripping metadata
- Reduce behavioral differences between GD and Imagick

### Proposed implementation

Intervention/Image already supports a `strip` option for Imagick encoders.
Glide only needs to forward the strip parameter from request parameters to the encoder options.

In `src/Api/Encoder.php`, inside the `run()` method, Glide could map the request parameter to Intervention/Image:

```php
if ($this->params['strip'] ?? false) {
    $encoderOptions['strip'] = true;
}
```
This does not add new image-processing logic to Glide; it simply exposes an existing Intervention/Image feature.

### Related note

More generally, I think it could also make sense to open a PR on Intervention/Image itself to further uniformize the default behavior between GD and Imagick, so metadata handling is consistent regardless of the underlying driver (cf https://github.com/Intervention/image/issues/1471).

I’d be happy to open a PR if this change sounds acceptable 🙂

Thanks for your work on Glide!

lazerg:
Glide never forwards a `strip` option to Intervention Image, so EXIF handling depends entirely on the driver — GD always drops metadata while Imagick always keeps it, with no way for a caller to control it. This lets the `driver` config accept an array of options (in addition to the existing driver name string), so `strip` and other `ImageManager` options can be set when configuring the server.

Fixes #445

### DIFF
--- a/docs/4.0/config/image-driver.md
+++ b/docs/4.0/config/image-driver.md
@@ -22,3 +22,20 @@ $server = \League\Glide\ServerFactory::create([
     'driver' => \Intervention\Image\Drivers\Vips\Driver::class,
 ]);
 ~~~
+
+## Driver options
+
+The `driver` option also accepts an array, letting you pass additional options through to Intervention Image's `ImageManager`, such as stripping EXIF metadata from output images.
+
+~~~ php
+<?php
+
+$server = \League\Glide\ServerFactory::create([
+    'driver' => [
+        'driver' => 'imagick',
+        'strip' => true,
+    ],
+]);
+~~~
+
+See the [Intervention Image driver configuration docs](https://image.intervention.io/v4/getting-started/configuration-drivers) for the full list of available options.
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -14,7 +14,6 @@ class Api implements ApiInterface
         'p', // preset
         'q', // quality
         'fm', // format
-        'strip', // strip metadata
         's', // signature
     ];
 
--- a/src/Api/Api.php
+++ b/src/Api/Api.php
@@ -14,6 +14,7 @@ class Api implements ApiInterface
         'p', // preset
         'q', // quality
         'fm', // format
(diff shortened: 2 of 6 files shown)


ADmad:
I don't think allowing `strip` as an Api (query string) parameter is necessary.

The `strip` option can be set for the [image manager](https://image.intervention.io/v4/getting-started/configuration-drivers) instance too. So we can add it as a `ServerFactory` option instead and use it when creating the `ImageManager` instance. 

Currently the `driver` config for `ServerFactory` can only be a string, we can allow it to be an array too thereby allowing to set other options for the ImageManager also.

lazerg:
Good call, that's a cleaner place for this. Pushed a commit that drops the `strip` API param and instead lets `driver` accept an array so `strip` (and other ImageManager options) can be set when configuring the server.

